PR #26, opened 2019-12-24
Correct the location name

jenting:
The backup-location and snapshot-location should be the it's name,
not the provider name. And the provider name should be it's name,
not the name default.

This change enhances after helm install, use `velero backup-location get`
and `velero snapshot-location get`, the NAME and PROVIDER fields are the
user's input.

Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.2.0
 description: A Helm chart for velero
 name: velero
-version: 2.7.9
+version: 2.7.10
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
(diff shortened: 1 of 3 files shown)


charts/velero/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.2.0
 description: A Helm chart for velero
 name: velero
-version: 2.7.5
```
nrb: Since #27 got in first, I think the version needs to be bumped to 2.7.7

charts/velero/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.2.0
 description: A Helm chart for velero
 name: velero
-version: 2.7.6
+version: 2.7.7
```
nrb: Sorry, looks like this is still causing a merge conflict :(

jenting:
@nrb Version bumped ;)

PR #33, opened 2020-01-09
RBAC: decouple cluster role binding from namespaced role/rolebinding.

desaintmartin:
Allows to have a namespaced Velero without cluster admin access or with externally setup ClusterRole.


carlisia:
@desaintmartin please rebase when you have a chance so we can merge after the second review!

desaintmartin:
Done!

PR #35, opened 2020-01-10
update GH actions

cpanato:
- update the GH actions

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,6 @@ jobs:
           helm repo add incubator https://kubernetes-charts-incubator.storage.googleapis.com/
 
       - name: Run chart-releaser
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -33,6 +33,6 @@ jobs:
           helm repo add incubator https://kubernetes-charts-incubator.storage.googleapis.com/
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.0.0-alpha.1
+        uses: helm/chart-releaser-action@v1.0.0-alpha.2
```
carlisia: Should this be `alpha.3`?

cpanato: for this, the `alpha.2` is the latest one, as we can see here: https://github.com/helm/chart-releaser-action/releases

carlisia:
@nrb PTAL when you have a chance.

PR #41, opened 2020-01-19
Control securityContext from values.yaml

### LINKED ISSUE #34: Add annotations on ServiceAccount to support IAM Roles for Service Accounts
Velero supports IAM Roles for Service Accounts as of v1.2.0: https://github.com/vmware-tanzu/velero/pull/1962

To actually use these the Helm chart needs to support custom `annotations` on the `ServiceAccount` resource, so a role annotation can be specified at install time. For example:
```
--set serviceAccount.server.annotations."eks\.amazonaws\.com/role-arn"=arn:aws:iam::123456789012:role/MyRoleName
```

As velero runs as `nobody`, [the pod will also need suitable permissions](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8) to read the projected token file:
```
securityContext:
  fsGroup: 65534
```

shaikatz:
Closes #34

### DIFF
--- a/charts/velero/templates/deployment.yaml
+++ b/charts/velero/templates/deployment.yaml
@@ -122,6 +122,8 @@ spec:
       initContainers:
         {{- toYaml .Values.initContainers | nindent 8 }}
 {{- end }}
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
       volumes:
         {{- if .Values.credentials.useSecret }}
         - name: cloud-credentials
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.2.0
 description: A Helm chart for velero
 name: velero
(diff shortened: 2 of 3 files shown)


charts/velero/templates/deployment.yaml
```diff
@@ -122,6 +122,8 @@ spec:
       initContainers:
         {{- toYaml .Values.initContainers | nindent 8 }}
 {{- end }}
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
```
carlisia: Thanks for adding this.  It would also be nice to add this to the restic DaemonSet as well, for parity.

This and a rebase and it should be good to go!

jenting:
I think #65 had fixed this issue, this PR could be closed.

nrb:
Looks like Jenting is correct - the other PR added the security context to both the deployment and the restic daemonset, so I'm going to close this PR out. Thank you for your efforts, @shaikatz!

PR #60, opened 2020-02-12
Initial release of vsphere-cpi chart to vmware-tanzu repo

dvonthenen:
This is chart is an exact copy of the vsphere-cpi chart in the helm/charts repo located: https://github.com/helm/charts/tree/master/stable/vsphere-cpi

I will be maintaining and keeping these two charts in sync going forward. I wasn't able to find any guidelines about chart submission, but if there are any feedback about what is currently there or what requirements are needed to get this chart accepted into the repo, feedback is welcome!

### DIFF
--- a/charts/vsphere-cpi/.helmignore
+++ b/charts/vsphere-cpi/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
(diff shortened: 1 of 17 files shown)


dvonthenen:
hey @nrb! Yes, I will definitely take on the responsibility for the chart's review and maintenance. If we could add @andrewsykim and @frapposelli that would be awesome. They are in the OWNERS file that was carried over from helm/charts repo... not sure if the privileges for approve/review/etc will carry over because of that file.

Also, I notice that the `Run chart-testing (install)` fails. I'm not sure how we could go about testing this because it requires a real vSphere environment along with a user-defined config (that could be based on that test environment). It seems like the long term we should probably figure something out to make that work (maybe use the VMC environment we have). For the short term, we could disable that specific test for the vSphere CPI, or just the test fail and ignore that particular error.

nrb:
Ok, I'll get them added to the chart team. We don't have any support for OWNERS files on this repo yet, but as its growing, we probably need to get that set up.

And I noticed the failure too, and figured it was because no vSphere environment was available. I'll see if I can mark it as passing for the moment, and we can see what's necessary to get a vSphere env hooked up for PRs. Right now we're using GitHub Actions with their own runners, and I've not explored setting up a custom runner yet.

nrb:
@dvonthenen Can you add the CI directory that at least populates some values for now, even if we don't have a full test environment quite yet?

carlisia:
@dvonthenen this seems to have gone stale. I'm going to close it but it can be reopened if needed.

Issue #7, opened 2019-12-12
Update Velero Helm chart for compatibility with Helm 3


carlisia:
Please see the original issue reported by @joschi36: 

https://github.com/helm/charts/issues/19387

carlisia:
Blocked by: https://github.com/vmware-tanzu/helm-charts/issues/2

carlisia:
No longer blocked.

wesmiles:
Hi, Is there a view to how long it will take to support Helm 3? Thanks

carlisia:
@cpanato would you have the bandwidth to tackle this in the near future?

cpanato:
yes @carlisia

marvinf95:
Is there any update about this issue? :)

branttaylor:
@cpanato @carlisia is there any estimate when Velero might support Helm v3? I know several other people have asked, but there has been no response so far.

vitobotta:
Hi @carlisia does the helm chart still support helm 2 only? Thanks

carlisia:
Hey @branttaylor, @wesmiles and @vitobotta, a PR review for this is wip: https://github.com/vmware-tanzu/helm-charts/pull/81

vitobotta:
Thanks @carlisia , hopefully it will be merged soon.

PR #81, opened 2020-03-15
Make CRDs Helm 3 compatiable

yurinnick:
- Remove `"helm.sh/hook": crd-install` from `crds/*.yaml`
- Copy Helm v2 CRDs to `templates/crds`, gated by `installCRDs` value

Based on [argo-cd](https://github.com/argoproj/argo-helm/tree/master/charts/argo-cd) helm chart.

Closes https://github.com/vmware-tanzu/helm-charts/issues/7.

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.3.1
 description: A Helm chart for velero
 name: velero
-version: 2.9.6
+version: 2.9.7
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/ci/test-values.yaml
+++ b/charts/velero/ci/test-values.yaml
@@ -1,3 +1,5 @@
+installCRDs: false
+
 # Set provider name and backup storage location bucket name
 configuration:
   provider: aws
--- a/charts/velero/crds/backups.yaml
+++ b/charts/velero/crds/backups.yaml
@@ -5,7 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: velero
   annotations:
-    "helm.sh/hook": crd-install
     "helm.sh/hook-delete-policy": "before-hook-creation"
 spec:
   group: velero.io
--- a/charts/velero/crds/backupstoragelocations.yaml
+++ b/charts/velero/crds/backupstoragelocations.yaml
@@ -5,7 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": crd-install
     "helm.sh/hook-delete-policy": "before-hook-creation"
 spec:
   group: velero.io
--- a/charts/velero/crds/deletebackuprequests.yaml
+++ b/charts/velero/crds/deletebackuprequests.yaml
@@ -5,7 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": crd-install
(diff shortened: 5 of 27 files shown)


yurinnick:
Suppose to fix CRDs creation for Helm 3 #7

jenting:
I was wondering if it possible we could do removing `"helm.sh/hook": crd-install` from `crds/*.yaml` without copy duplicated CRDs to `templates/crds/*.yaml` to make it compatible on both helmv2 and helmv3.

BTW, I think set `installCRDs=false` when helm install would cause velero server install failed because when velero server startup, velero tries to access the velero CRDs resources.

yurinnick:
I don't this there is a way to keep backward compatibility without having a copy of CRDs with a `crd-install` hook. Why Helms team decided to deprecate this hook in such a breaking way is beyond me.

For Helm v3 `installCRDs=false` suppose to keep templated CRDs with `crd-install` hook from installing, so they won't break a deployment.

yurinnick:
@carlisia this seems to be the critical functionality for many people using Helm v3. Is there any chance to get it merge in nearest future?

carlisia:
Hey @yurinnick I missed this PR! I will review it later today, and thank you!

charts/velero/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.3.0
 description: A Helm chart for velero
 name: velero
-version: 2.9.3
+version: 2.9.4
```
cpanato: please rebase your PR and bump the chart version

yurinnick:
Blocked by #85

carlisia:
I'm still testing this. So far:

- getting the `manifest_sorter.go:192: info: skipping unknown hook: "crd-install"`. Investigating if we could make it so we don't output this, maybe change how the hook is handled.
- the velero deployment is not being created. I'm looking to see if it's something on my end.

yurinnick:
@carlisia by default installation uses Helm v2 approach of imstalling CRDs from templates. To enable Helm v3 "mode", please set `installCRDs: false` in `values.yaml`. Should I enable Helm v3 "mode" by default?

carlisia:
Ah, thank you! Let me review the whole thing and see what changes if any we should make.

yurinnick:
Okay, so with `ct` using Helm v3 by default I disabled templated CRDs installation for CI. Also, I gated `cleanup-cdrs` with all other Helm v2 CRDs templates, so it won't explode CI builds on cleanup.

carlisia:
Should `cleanUpCRDs` exist in the `values.yaml`?

yurinnick:
Well, for Helm v3 deployments including CI it doesn't matter now and it worked for Helm v2 before, so... I guess it's fine 😃 . I will add default value for in into `values.yaml` though.

carlisia:
I'm able to install using Helm 3. But my restore is not restoring as I expected. Investigating. 

Still need to check compatibility with Helm 2.

carlisia:
- [ ] backup w/ Heml 3
- [ ] restore w/ Heml 3
- [ ] backup w/ Heml 2
- [ ] restore w/ Heml 2

yurinnick:
Could you provide steps to reproduce as well as error messages if you got any?

carlisia:
No, I got it to pass (see the checkboxes above). Today I'll test the chart with Helm 2.

gadelkareem:
@carlisia IMHO if someone needs Helm 2 then they can just use an older chart with a diff version in the values

carlisia:
@nrb / @cpanato / @skriss PTAL when you have a chance.

I volunteer to add a new PR with an update to the documentation.

yurinnick:
@nrb @cpanato could you review please?

cprivite:
Does README.md need to be updated to tell people to supply installCRDs: false for helm 3 installs?

carlisia:
@cprivite yes, that's probably a good idea. Would you like to add it?

PR #94, opened 2020-04-09
Fixing toYaml generation

limed:
This fixes the `toYaml` generation of the template and resolves issue #93, it appears that the release of version 2.9.9 of the chart causes issue with the template generation and the template doesn't seem to render if you use `{{ toYaml .config }}`

This fix actually renders the template locally after removing the `.config`

Fixes https://github.com/vmware-tanzu/helm-charts/issues/93.

PR #99, opened 2020-04-13
Add support for secret env vars

dunglas:
This PR adds a new value `credentials.extraEnvVars` allowing to store data in a secret and expose it to the container through an environment variable.
This is necessary for the [Digital Ocean plugin](https://github.com/digitalocean/velero-plugin) and maybe other.

I tested it (and it works). This PR also includes #94 (because without this PR, the chart is broken). I'll rebase when #94 will be merged.

Thanks for this great tool by the way!

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.3.1
 description: A Helm chart for velero
 name: velero
-version: 2.9.14
+version: 2.9.15
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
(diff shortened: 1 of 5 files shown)


skriss:
@carlisia yeah, I think it probably should, since the secret is mounted in there too so presumably any extra env vars apply there as well.

carlisia:
Ok @dunglas, please also add it to the `restic-daemonset` template when you have a chance.

dunglas:
@carlisia done! Thanks for the review.

qRoC:
Hello.

>  variables such as "DIGITALOCEAN_TOKEN: <your-key>".

What is `<your-key>`?

For configuration:

```yaml
credentials:
  existingSecret: velero-secrets
  extraEnvVars:
    DIGITALOCEAN_TOKEN: 'hello'
```

Need generate 2 secrets: `velero-secrets` and `velero`?

Because is generate:

```yaml
- name: DIGITALOCEAN_TOKEN
   valueFrom:
      secretKeyRef:
      name: velero
      key: DIGITALOCEAN_TOKEN
```

mshivanna:
In deployment.yaml file on line 129 under secretKeyRef `name: {{ include "velero.fullname" $ }}` should have been  `name: {{ include "velero.secretName" $ }}`

PR #104, opened 2020-05-10
Update Readme for Helm3

mickkael:
set installCRDs to false to avoid the unnecessary warnings output when installing with helm 3.
as proposed here : https://github.com/vmware-tanzu/helm-charts/pull/81#issuecomment-615406209


carlisia:
Hey @mickkael thank you for this.

We've made some changes on PR https://github.com/vmware-tanzu/helm-charts/pull/125 so now we no longer get the warnings when using Helm3. Thanks for the PR but in light of that change, I'm going to close this.

image

PR #109, opened 2020-05-25
Helm 2,3 supports CRD at the same time

tomahawk28:
Hi, 
This is the same PR we open for supporting helm 2 & 3 at the same time what we have. 
https://github.com/giantswarm/velero-app/pull/3

After patching this, the helm would use CRD in one directory.


tomahawk28:
Would you mind to take a look? cc> @carlisia, @cpanato

charts/velero/templates/backupstoragelocation.yaml
```diff
@@ -3,6 +3,9 @@ apiVersion: velero.io/v1
 kind: BackupStorageLocation
 metadata:
   name: {{ include "velero.backupStorageLocation.name" . }}
+  annotations:
+    "helm.sh/hook": post-install
```
jenting, line 7: A question, why this not pre-install but post-install instead? [1]

charts/velero/templates/volumesnapshotlocation.yaml
```diff
@@ -3,6 +3,9 @@ apiVersion: velero.io/v1
 kind: VolumeSnapshotLocation
 metadata:
   name: {{ include "velero.volumeSnapshotLocation.name" . }}
+  annotations:
+    "helm.sh/hook": post-install
```
jenting, line 7: [1]

charts/velero/templates/crds/cleanup-crds.yaml
```diff
@@ -1,39 +0,0 @@
-# This job is meant primarily for cleaning up on CI systems.
```
jenting, line 1: I think it would be good to keep this file since it supports delete CRDs.

carlisia: Yes, I agree. We don't need to keep it inside the `crds` folder I don't think.

For reference:
https://github.com/vmware-tanzu/helm-charts/blob/efbcab45316125c5760a63cc75f2343879cf3fc0/charts/velero/ci/test-values.yaml#L26

carlisia:
I'm testing this PR today.

charts/velero/templates/crds/cleanup-crds.yaml
tomahawk28: Good point, I revert it and placed in under `templates`

tomahawk28:
@jenting @carlisia PTAL!

sebastien-prudhomme:
@tomahawk28 for Helm2, this commit breaks tools such are Argo CD that are using some king of "helm template" without using Tiller service.

tom-hadlaw-hs:
> @tomahawk28 for Helm2, this commit breaks tools such are Argo CD that are using some king of "helm template" without using Tiller service.

I'm having the same issue but for Helm3

PR #110, opened 2020-05-26
Bump Velero v1.4.0 and plugin v1.1.0

carlisia:


charts/velero/values.yaml
```diff
@@ -6,8 +6,8 @@
 # enabling restic). Required.
 image:
   repository: velero/velero
-  tag: v1.3.1
-  digest: sha256:0c74f1d552ef25a4227e582f4c0e6b3db3402abe196595ee9442ceeb43b99696
+  tag: v1.4.0
+  digest: sha256:d238835e151cec91c6a811fe3a89a66d3231d9f64d09e5f3c49552672d271f38
```
skriss, line 10: Just wondering if we should leave "digest" empty but comment how to use it and that it takes precedence over `tag` if it's specified.

carlisia: Yes, I think that's a good idea.

PR #117, opened 2020-06-11
Add log configuration to Restic

### LINKED ISSUE #116: Restic daemon set does not honour log configuration
The Restic daemon set does not use values about log in `.Values.configuration`.  
It would be convenient to debug.

The fix is to modify `restic-daemonset.yaml` to include `logLevel` and `logFormat` as in the `deployment.yaml` file.

fredgate:
Fix #116


cpanato:
@fredgate can you rebase your PR and resolve the conflict? thanks!

PR #119, opened 2020-06-22
fix: Restic daemonset SA always set, like velero deployment

### LINKED ISSUE #118: Restic serviceAccountName setting not used with existing service accounts
The Restic daemonset should set the `serviceAccountName` regardless of the service account creation setting. This would allow using eg. pre-created IAM role-bound service accounts as with the Velero server deployment.

Instead of:
```yaml
{{- if .Values.serviceAccount.server.create }}
serviceAccountName: {{ include "velero.serverServiceAccount" . }}
{{- end }}
``` 

It should not filter:
```yaml
serviceAccountName: {{ include "velero.serverServiceAccount" . }}
```

kyounger:
Signed-off-by: Kenny Younger <kyounger@gmail.com>

I don't think there is a reason that the SA should be gated. Existing service accounts should be supported.

Fixes #118

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.4.0
 description: A Helm chart for velero
 name: velero
-version: 2.12.10
+version: 2.12.11
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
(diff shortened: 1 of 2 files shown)


kyounger:
Sorry, not familiar with the github actions step you use to lint and why it might be failing. Best I can tell the syntax of the action is incorrect.

jenting:
You missed bumping the helm chart version in file _Chart.yaml._

kyounger:
@carlisia hope that latest push of mine is sufficient.

carlisia:
@kyounger sorry, we've been very active merging PRs today. One more version bum please, yours is next!

kyounger:
Oops. I think I broke something with my latest rebase. Fixing that and will bump version, too.

kyounger:
Alrighty. I think that does it!

PR #125, opened 2020-07-23
helm 2,3 CI test + Upgrade CRDs to v1.4

jenting:
1. change target branch from master to main
2. add helm2 & helm3 CI test
3. bump the CRDs manifests sync with velero `v1.4.0`

then #104 is no more needed.

Signed-off-by: JenTing Hsiao <jenting.hsiao@suse.com>

PR #126, opened 2020-07-25
Fix Helm 2 CRD generation

MrFreezeex:
Signed-off-by: Arthur Outhenin-Chalandre <arthur@cri.epita.fr>

I have this issue when I install the latest velero helm chart with ArgoCD:
```
rpc error: code = Unknown desc = Failed to unmarshal manifest: error unmarshaling JSON: Object 'Kind' is missing in '{"apiVersion":"apiextensions.k8s.io/v1beta1"}'
```

ArgoCD recognizes the velero chart as an helm2 chart and therefore uses the helm2 binary for it.
I tested and I saw that the line `apiVersion: apiextensions.k8s.io/v1beta1` is not indented correctly with helm2 (the output in helm3 is correct). This PR is fixing this indentation issue.

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.4.0
 description: A Helm chart for velero
 name: velero
-version: 2.12.2
+version: 2.12.3
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.4.0
 description: A Helm chart for velero
 name: velero
-version: 2.12.1
+version: 2.12.2
 home: https://github.com/vmware-tanzu/velero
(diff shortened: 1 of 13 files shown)


MrFreezeex:
I bumped the chart version, should fix the CI.

MrFreezeex:
I removed the leading whitespace on CRD that was used in the #125 to fix this issue.

nrb:
@MrFreezeex CI had some intermittent failures with the AWS plugin installing, so I re-ran it. I don't think the whitespace removal is 100% necessary, but doesn't look like it hurts either.

MrFreezeex:
> @MrFreezeex CI had some intermittent failures with the AWS plugin installing, so I re-ran it. I don't think the whitespace removal is 100% necessary, but doesn't look like it hurts either.

Yes I figured It out, I think It was kind of a hack to fix the chart in the previous PR.

jenting:
> > @MrFreezeex CI had some intermittent failures with the AWS plugin installing, so I re-ran it. I don't think the whitespace removal is 100% necessary, but doesn't look like it hurts either.
> 
> Yes I figured It out by testing with kubectl, I think the leading line was kind of a hack to fix the chart in the previous PR (having the first empty line indented in every CRD).

Well, it's not hacking but directly copy from https://github.com/vmware-tanzu/velero/blob/v1.4.0/pkg/generated/crds/manifests/velero.io_backups.yaml, it includes the first empty line on all CRDs.

PR #130, opened 2020-07-27
Fix set-file credentials

### LINKED ISSUE #103: Helm install documentation has no info about how to specify the --set-file key
When you run helm install with the --set-file argument, you need to provide a key to which the file you're passing in is mapped to. Unfortunately, the current helm-charts/charts/velero/README.md  doesn't give you any clues about what that should be. 

This was handed in the past via the main velero documentation by the docs created in this issue: https://github.com/vmware-tanzu/velero/issues/1533 

Unfortunately, the documentation has been revised since version 1.1.0 and that page doesn't exist anymore. Thus we've lost the only clue about how to properly format the --set-file argument and now the currently provided helm install block in the readme isn't valid or helpful.

The AWS plugin documentation is not helpful at all as it ignores the helm use case entirely.

I've only figured out how to fix this for the AWS plugin, so I can't really suggest a generic solution for the document, but I do know that for a new user doing a new install right now, this was a major blockage in being able to use the helm install of Velero whatsoever.

For passing in AWS credentials file that most walkthroughs have you create, here's what you want:

`--set-file credentials.secretContents.cloud=./credentials-velero`

If nothing else, hopefully this issue will save someone else hours of debugging.

carlisia:
Fixed #103.

Signed-off-by: Carlisia <carlisia@vmware.com>

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.4.0
 description: A Helm chart for velero
 name: velero
-version: 2.12.4
+version: 2.12.6
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/README.md
+++ b/charts/velero/README.md
@@ -37,7 +37,7 @@ Specify the necessary values using the --set key=value[,key=value] argument to h
 
 ```bash
 helm install vmware-tanzu/velero --namespace <YOUR NAMESPACE> \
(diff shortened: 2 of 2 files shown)


ashish-amarnath:
Actually quick question.. does the README file update also need a helm chart version bump?

carlisia:
I know....! Yes.

charts/velero/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.4.0
 description: A Helm chart for velero
 name: velero
-version: 2.12.5
+version: 2.12.6
```
ashish-amarnath, line 5: Is this change required even for an update to the README file?

PR #146, opened 2020-08-17
[velero] A one time job to install/upgrade/delete CRDs

jenting:
#### Special notes for your reviewer:

Use a one time job to install/upgrade/delete CRDs with the corresponding service account and RBAC.

With this PR:
- we don't have to manage the CRDs manifests in code and check it in sync with https://github.com/vmware-tanzu/velero repo
- we could make sure that the helm upgrade will upgrade the CRDs, please ref to https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

fixes #44 

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[velero]`)

Issue #44, opened 2020-01-23
Drop CRDs manifest but install CRDs by velero install


jenting:
Right now, the velero helm chart CRD manifest is not in-sync with velero repo.
Also the command `velero install --crds-only` supports install CRDs.

I'm wondering that is it better to drop helm-chart CRDs manifest but install the CRDs by `veleror install --crds-only`?
This way makes maintaining the CRDs from the single source and does not need to address sync CRD manifest from velero repo to velero helm chart.

Are there any other cons I did not think about? Thanks.

carlisia:
@cpanato do you have any opinion?

I'm going to add this to next's week community meeting agenda.

carlisia:
https://hackmd.io/Jq6F5zqZR7S80CeDWUklkA

cpanato:
I saw other projects that don't install the CRDs as part of the helm installation, but I don't recall which ones were :(

I think this is a good idea but we need to make sure we output the additional commands we need to apply after we install the chart.

jenting:
I was work on this previously, if this is a good approach, I could help on this 😉

nrb:
I've also seen other projects do this, due to the way Helm's designed.

It comes down to Helm wanting to manage a release's lifecycle, while a CRD by definition must live much longer than any one release. Helm has tried to fix it in several ways, and so far I haven't seen any of them be particularly reliable in Helm 2; I've not done any experimenting in Helm 3 to know if the situation has improved. 

I do agree that maintaining one source of definitions would be good, because having them get out of sync is not good for anyone.

carlisia:
Just something to think about: we have the intention of dropping the `velero install` command: https://github.com/vmware-tanzu/velero/pull/2202

jenting:
My idea on how to achieve this is:
1. Run a Job which executes "velero install --crds-only" with helm pre-install hook to make sure that the CRDs are installed.
2. The tests resources are installed with helm post-install hook.

However, after testing on helm3, it seems like the helm would not waits the Job executes completed then install the other resources. It means that right now I cannot find a way to achieve this.

BTW, I found two problems in current helm charts:
1. The CRDs are out-of-sync compares to current Velero version v1.4.2
2. The helm would not handles update existing CRDs when helm upgrade. Ref to https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

carlisia:
We'd need to generate the CRDs originally already containing these annotations right:

[...]

PR #171, opened 2020-10-14
[velero] Add enable/disable option for helm hooks

### LINKED ISSUE #158: Backup is not creating with failed validation error on ArgoCD
**What steps did you take and what happened:**
 I have scheduled mybackup with 2 hours

**What did you expect to happen:**
The backup process should be completed and i shud be able to see the data in my s3 bucket

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other pastebin is fine.)
```apiVersion: velero.io/v1
kind: Backup
metadata:
  annotations:
    velero.io/source-cluster-k8s-gitversion: v1.17.9-eks-4c6976
    velero.io/source-cluster-k8s-major-version: '1'
    velero.io/source-cluster-k8s-minor-version: 17+
  creationTimestamp: '2020-09-08T13:00:09Z'
  generation: 2
  labels:
    app.kubernetes.io/instance: velero
    app.kubernetes.io/managed-by: Tiller
    app.kubernetes.io/name: velero
    helm.sh/chart: velero-2.12.16
    velero.io/schedule-name: velero-hourly-backup
    velero.io/storage-location: aws
  name: velero-hourly-backup-20200908130009
  namespace: velero
  resourceVersion: '60355'
  selfLink: >-
    /apis/velero.io/v1/namespaces/velero/backups/velero-hourly-backup-20200908130009
  uid: 411b3648-a71e-400a-92cf-a638cc3d84aa
spec:
  hooks: {}
  includeClusterResources: true
  includedNamespaces:
    - '*'
  includedResources:
    - '*'
  snapshotVolumes: true
  storageLocation: aws
  ttl: 720h0m0s
status:
  expiration: '2020-10-08T13:00:09Z'
  formatVersion: 1.1.0
  phase: FailedValidation
  validationErrors:
    - >-
      a BackupStorageLocation CRD with the name specified in the backup spec
      needs to be created before this backup can be executed. Error:
      backupstoragelocation.velero.io "aws" not found
  version: 1```

[Miscellaneous information that will assist in solving the issue.]


**Environment:**

- helm version (use `helm version`): 
- helm chart version and app version: 1.4.2 and helm version 2.12.16
  - helm2 (use `helm list`):
  - helm3 (use `helm list -n <YOUR NAMESPACE>`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version: eks version 1.18
- Cloud provider or hardware configuration: aws
- OS (e.g. from `/etc/os-release`): linux

### PR DESCRIPTION
**Pull request by igorcezar**

Create an option in values to disable helm hooks used in CRDs (for Helm v2) and templates manifest (like BackupStorageLocation), introduced in #109 and #153. 

With this change it's possible to install this chart using ArgoCD without any problems, setting the new option to false:

```
enableHelmHooks: false
```

It's possible to disable only CRD hooks or template hooks, in case you don't want to disable both.

The default option is true, which enables all helm hooks.

Helm hooks in crds folder, for Helm v3, are untouched, because they can't be templated.

Fix #158 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -72,8 +72,9 @@ metrics:
 # Install CRDs as a templates. Enabled by default.
 installCRDs: true
 
-# Enable/disable helm hooks annotations for CRDs (Helm v2) and templates
-# You should disable this if using a deploy tool that doesn't support helm hooks
+# Enable/disable all helm hooks annotations
+# You should disable this if using a deploy tool that doesn't support helm hooks,
+# such as ArgoCD
 enableHelmHooks: true
 
 ##
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -72,11 +72,9 @@ metrics:
 # Install CRDs as a templates. Enabled by default.
 installCRDs: true
 
-# Enable/disable helm hooks annotations for Helm v2 in CRDs and/or templates
+# Enable/disable helm hooks annotations for CRDs (Helm v2) and templates
 # You should disable this if using a deploy tool that doesn't support helm hooks
-enableHelmHooks:
-  crds: true
-  templates: true
+enableHelmHooks: true
 
 ##
 ## End of deployment-related settings.
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -72,6 +72,12 @@ metrics:
 # Install CRDs as a templates. Enabled by default.
 installCRDs: true
 
+# Enable/disable helm hooks annotations for Helm v2 in CRDs and/or templates
+# You should disable this if using a deploy tool that doesn't support helm hooks
+enableHelmHooks:
+  crds: true
+  templates: true
+
 ##
 ## End of deployment-related settings.
 ##
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.5.1
+appVersion: 1.5.2
 description: A Helm chart for velero
 name: velero
-version: 2.13.4
+version: 2.13.7
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.5.1
 description: A Helm chart for velero
 name: velero
-version: 2.13.3
+version: 2.13.4
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/crds/backups.yaml
+++ b/charts/velero/crds/backups.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: velero
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: backups.velero.io
 spec:
--- a/charts/velero/crds/backupstoragelocations.yaml
+++ b/charts/velero/crds/backupstoragelocations.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: backupstoragelocations.velero.io
 spec:
--- a/charts/velero/crds/deletebackuprequests.yaml
+++ b/charts/velero/crds/deletebackuprequests.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: deletebackuprequests.velero.io
 spec:
--- a/charts/velero/crds/downloadrequests.yaml
+++ b/charts/velero/crds/downloadrequests.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: downloadrequests.velero.io
 spec:
--- a/charts/velero/crds/podvolumebackups.yaml
+++ b/charts/velero/crds/podvolumebackups.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: podvolumebackups.velero.io
 spec:
--- a/charts/velero/crds/podvolumerestores.yaml
+++ b/charts/velero/crds/podvolumerestores.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "velero"
   annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-delete-policy": "before-hook-creation"
     controller-gen.kubebuilder.io/version: v0.3.0
   name: podvolumerestores.velero.io
 spec:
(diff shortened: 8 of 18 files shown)

### CODE REVIEW

**Review by cidsasaki — DISMISSED** [review]
Works as expected! Thanks for the correction.

**Comment by cidsasaki** [comment]
Resolves to #146 / #44 / #158

File: charts/velero/values.yaml [hunk 1]
```diff
@@ -72,6 +72,12 @@ metrics:
 # Install CRDs as a templates. Enabled by default.
 installCRDs: true
 
+# Enable/disable helm hooks annotations for Helm v2 in CRDs and/or templates
```
jenting, line 75: I think it's not specific on helm v2 only.

charts/velero/values.yaml
```diff
@@ -72,6 +72,12 @@ metrics:
...
+# You should disable this if using a deploy tool that doesn't support helm hooks
+enableHelmHooks:
+  crds: true
+  templates: true
```
jenting, line 79: I'm thinking if we could use only one variable to cover all the scenarios?
For ArgoCD, what's the `enableHelmHooks.crds` and `enableHelmHooks.templates` would be?

jenting:
@simran3695, @fabianmet could you please take a look a this 🙇 ? It seems to fix #158.

jenting:
> Resolves to #146 / #44

thanks, but I don't think it resolves the #44.

charts/velero/values.yaml
igorcezar: ArgoCD uses only Helm v2 until v1.4. Starting in v1.5 it has both v2 and v3, using v3 when Chart.yaml apiVersion is v2.

charts/velero/values.yaml
igorcezar: We could. At first I thought about use only one variable for all, like "enableHelmHooks: true", but I made it that way to cover any scenarios where disabling hooks on CRDs it's necessary but not on templates, although I don't have any in mind. 

ArgoCD has no problem with helm hooks on templates from v1.3, so it works with or without helm hooks on them.

One variable for all make it simpler, and is better, if there is really no use for keeping separated.

jenting: @simran3695, @fabianmet could you please provide the ArgoCD version you used?
I can't reproduce #158 on ArgoCD version 1.7.

charts/velero/values.yaml
jenting: okay, since we're using `apiVersion: v1` https://github.com/vmware-tanzu/helm-charts/blob/c90e413/charts/velero/Chart.yaml#L1, so the ArgoCD would use helm v2 behavior even we use ArgoCD version >= 1.5, right?

charts/velero/values.yaml
jenting: > We could. At first I thought about use only one variable for all, like "enableHelmHooks: true", but I made it that way to cover any scenarios where disabling hooks on CRDs it's necessary but not on templates, although I don't have any in mind.
> 
> ArgoCD has no problem with helm hooks on templates from v1.3, so it works with or without helm hooks on them.
> 
> One variable for all make it simpler, and is better, if there is really no use for keeping separated.

I can't think of other scenarios, for now, but using a single variable makes it simpler to understand.

charts/velero/values.yaml
igorcezar: That’s right!

charts/velero/values.yaml
igorcezar: > I can't think of other scenarios, for now, but using a single variable makes it simpler to understand.

Shall I change this to a single variable then? 

>@simran3695, @fabianmet could you please provide the ArgoCD version you used?
I can't reproduce #158 on ArgoCD version 1.7.

I saw the same problem that @simran3695 describe in #158 on version 1.5, where CRDs with PreSync keep syncing forever. Like @fabianmet [explained](https://github.com/vmware-tanzu/helm-charts/issues/158#issuecomment-692009835), it's a problem on how ArgoCD treats helm hooks, and when helm hooks are removed, all works fine.
For CRs like BSL and VSL I had no problems when using helm hooks/Post Sync on v1.5, but removing hooks also works fine (setting enableHelmHooks.templates: false)

jenting: Maybe use a single variable would be easier:
```
enableHelmHooks: true
```

fabianmet (commented):
The actual issue arises because if you install this chart with helm v3 the pre-install pre-sync hooks are also triggered on the CRD's in the CRD directory.

This causes the state to be dropped by ArgoCD because a hook has no state. It is a one off job.
The only way to fix this is either remove the hook notations in the CRD directory all together, or change them back to the helm hook for crd-install.

I understand you want this helm chart to keep the CRD's in line with your actual helm velero installation but this goes against what helm themselves say as explained [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations).

I tried the same approach as @igorcezar did in this PR but it did not solve the issue where the CRD's were installed but constantly in a bad state because of the pre-install pre-sync hooks still being present on the CRD's themselves.

Also its a shame but the CRD directory does not treat files as templates, so there is no way to properly disable helm hooks in the CRD files.

Having the option to see what changes in a CRD is really helpful for upgrade purposes. So im a big fan of having them available easily in git instead of having velero themselves generate it but that is my opinion.

charts/velero/values.yaml
igorcezar: Done! We can change this in the future if comes any scenario for that.

cidsasaki (approved):

charts/velero/values.yaml
igorcezar: > Not exactly, if you want to provide your own values.yaml for this chart for the helm install you need to build your own chart which has this one as a dependency. Which results in usually helm v3 being the actual engine.
This is because argocd requires the source of both helm chart and altered values.yaml to be in the same repository.

I use it that way, with velero chart as a dependency, and ArgoCD also uses Helm v2. For Helm v3 I have to change Chart.yaml apiVersion of the chart that I build (the one which imports velero chart) to v2, or force ArgoCD to use Helm v3 [as described here](https://argoproj.github.io/argo-cd/user-guide/helm/#helm-version).

igorcezar:
> The only way to fix this is either remove the hook notations in the CRD directory all together, or change them back to the helm hook for crd-install.

In this PR we remove hook notations from templates/crds.yaml and the CR's templates, which for ArgoCD solves the problem, since it uses Helm v2 if you don't change the chart apiVersion to v2. But for sure the best solution would be to also remove from CRD folder so ArgoCD can work even using Helm v3.

@jenting  the hook notations were included in CRD directory just for templates/crds.yaml could import the crds with the hooks, right?
Could we remove it from CRD folder (Helm v3) and enable only in crds.yaml for Helm v2?

jenting:
> > The only way to fix this is either remove the hook notations in the CRD directory all together, or change them back to the helm hook for crd-install.
> 
> In this PR we remove hook notations from templates/crds.yaml and the CR's templates, which for ArgoCD solves the problem, since it uses Helm v2 if you don't change the chart apiVersion to v2. But for sure the best solution would be to also remove from CRD folder so ArgoCD can work even using Helm v3.
> 
> @jenting the hook notations were included in CRD directory just for templates/crds.yaml could import the crds with the hooks, right?

I think so.

> Could we remove it from CRD folder (Helm v3) and enable only in crds.yaml for Helm v2?

Not sure what you'd like to remove? The helm annotation _or_ CRD folder?

igorcezar:
> Not sure what you'd like to remove? The helm annotation or CRD folder?

The helm annotation in all *.yaml of CRD folder. 
Keep annotation only for templates/crds.yaml file.

igorcezar:
I have pushed the changes I was talking about. 

With this, the chart works for ArgoCD with Helm v2 (setting  **enableHelmHooks: false**) and Helm v3 ( no matter how enableHelmHooks is set, but the preference is set to false, so that BSL/VSL/Schedule don't become Post Sync hooks).

Also works with **helm install** for both versions too (tested with helm v2.15.2 and v3.1.2), with enableHelmHooks: true (which is the default).

cidsasaki (approved):

jenting:
I want to play ArgoCD so I'll review this PR slowly.

charts/velero/values.yaml
```diff
@@ -72,6 +72,10 @@ metrics:
 # Install CRDs as a templates. Enabled by default.
 installCRDs: true
 
+# Enable/disable helm hooks annotations for CRDs (Helm v2) and templates
+# You should disable this if using a deploy tool that doesn't support helm hooks
```
jenting, line 76: 1. I don't think it's for helm2 only, these helm hooks also available on helm3
2. maybe we could mention the deploy tool ArgoCD to help people easier to understand when to disable helm hooks annotations.

jenting (changes requested):
@igorcezar could you please address the comments and fix the conflicts? thank you.

charts/velero/values.yaml
igorcezar: For CRDs it was helm2 only, because in helm3 CRDs can't be templated, but for all else it was for both v2 and v3.

jenting: > For CRDs it was helm2 only, because in helm3 CRDs can't be templated, but for all else it was for both v2 and v3.

```
helm template vmware-tanzu/velero --include-crds
```

igorcezar: Hooks are available for Helm3, but in this chart the option enableHelmHooks will only enable/disable helm hooks for helm2 CRDs (templates/crds.yaml) and for all other templates on both versions. 
I've made the changes in the comments and chart version.

jenting (approved):
LGTM, thank you

igorcezar:
Any news on this?

jenting:
> Any news on this?

@carlisia @cpanato PTAL

cpanato (approved):
lgtm, did not see any issues

thanks!

PR #205, opened 2021-01-21
[velero] Helm2 fix

cpanato:

The previous helm releases (<2.17) had the stable repository hardcoded, the problem is the URL does not exist anymore. 
So using the last helm v2 release 2.17.0 solves the issue.

We use the `chart-testing` to run all tests, in the helm side we are checking if we will release a final `chart-testing` for V2 or not, so in meantime, I've pushed the `chart-testing` image to my quay.io repo and we can use that and unblock the pipeline for v2

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.5.2
 description: A Helm chart for velero
 name: velero
-version: 2.14.5
+version: 2.14.6
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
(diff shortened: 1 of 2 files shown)


charts/velero/Chart.yaml
```diff
@@ -18,4 +18,3 @@ maintainers:
     email: brubakern@vmware.com
   - name: skriss
     email: krisss@vmware.com
-tillerVersion: ">=2.10.0"
```
jenting, line 21: Remove this won't harm helm v2?

cpanato: this is optional, and I dont think users are using helm v2 under 2.9 :)

cpanato: but removed that, because the chart-testing removed that from the lint schema, so if we keep the lint will fail

jenting (approved):

carlisia (approved):
Thank you!

PR #208, opened 2021-01-27
[velero] allow disabling schedule in value overrides

tareqhs:

In multi-layer environment where several values files are applied, it is very useful to allow disabling a schedule that is defined in a common values file.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.5.3
 description: A Helm chart for velero
 name: velero
-version: 2.14.7
+version: 2.14.8
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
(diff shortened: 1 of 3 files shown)


jenting:
Generally LGTM, but one question, can't we just set the "schedules: {}"?

tareqhs:
> Generally LGTM, but one question, can't we just set the "schedules: {}"?

setting a field as empty object does not affect its contents - it simply merges with no updates. probably you mean this:
```yaml
schedules: null
```
but this removes all the schedules and the goal is controlling individual schedules. i can follow on your line of thought and make the template support cases like this:
```yaml
schedules:
  myschedule: null
```
the template currently break on inputs like this. actually i like this approach as it spares the extra field.

jenting:
I see! You would like to disable some scheduling settings but want to keep it in `values.yaml`. So, in the feature, you might want to re-enable it, right?

jenting:
> I see! You would like to disable some scheduling settings but want to keep it in `values.yaml`. So, in the feature, you might want to re-enable it, right?

ping @tareqhs

jenting:
@tareqhs Is your use case still need this PR enhancement?

jenting:
@tareqhs I'd prefer to close this PR since I don't fully understand what's your use cases. Free free to reopen it once you have a chance to revisit it. :bow:

PR #214, opened 2021-02-03
[velero] - Change Release.Service to "Helm" for v3 support

jsmcnair:
Signed-off-by: John McNair <john.mcnair@yellowdog.co>

I'm using Grafana Tanka's Helm wrapper to install the velero chart. I was getting an error about a resource type not existing and noticed that the CRDs were not being deployed. I think you need to change the Release.Service to "Helm" for v3 support. I've done this and it worked for me. I have not tested with vanilla Helm though.

Feel free to ignore if I'm barking up the wrong tree! 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.5.3
 description: A Helm chart for velero
 name: velero
-version: 2.14.8
+version: 2.14.9
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/templates/crds.yaml
+++ b/charts/velero/templates/crds.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.installCRDs (eq .Release.Service "Tiller") }}
+{{- if and .Values.installCRDs (eq .Release.Service "Helm") }}
 {{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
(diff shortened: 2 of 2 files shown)


jenting:
@jsmcnair thanks for your contribution, please bump the Chart version to fix CI failure.

jsmcnair:
I updated the patch version to 2.14.9 but now it fails with:
```
Error: failed pre-install: warning: Hook pre-install velero/templates/crds.yaml failed: customresourcedefinitions.apiextensions.k8s.io "backups.velero.io" already exists
```
Did I do something wrong?

jenting:
Currently, velero helm-chart supports helmv3 already. So I'm not sure why the Grafana Tanka's Helm wrapper doesn't work.

jsmcnair:
I will investigate. 

Any thoughts on the linter error? It seems to be complaining that the CRDs are already installed.

jenting:
> I will investigate.
> 
> Any thoughts on the linter error? It seems to be complaining that the CRDs are already installed.

It failed at `helm upgrade` test, no idea for now.

jenting:
> I will investigate.
> 

Have the problem you encountered be resolved already?

mkilchhofer:
This fails because the CRDs already exists in the `/crds/` directory. Helm3 installs the CRDs inside this directory once. But after initial installation, helm does not do anything with these CRDs.

The magic for `/crds/` directory is that they get installed (only if absent) and helm then waits that they become available on the the API before applying the rest of the chart outside this `/crds/` dir. This is useful when a single chart installs both the definitions and an instance of that definition. `BackupStorageLocation` is such an example in this chart here (when `backupsEnabled` is set to true).

PR #215, opened 2021-02-10
[Velero] Add velero component label and update docs

codegold79:
Relates to changes made by [vmware-tanzu/velero PR3447](https://github.com/vmware-tanzu/velero/pull/3447). 
- add `component=velero` label
- remove `--generate-name` and use "velero" as name.
- update documentation

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/README.md
+++ b/charts/velero/README.md
@@ -69,7 +69,7 @@ helm install vmware-tanzu/velero --namespace <YOUR NAMESPACE> -f values.yaml --g
 If a value needs to be added or changed, you may do so with the `upgrade` command. An example:
 
 ```bash
-    helm upgrade vmware-tanzu/velero <RELEASE NAME> --namespace <YOUR NAMESPACE> --reuse-values --set configuration.provider=<NEW PROVIDER>
+helm upgrade vmware-tanzu/velero <RELEASE NAME> --namespace <YOUR NAMESPACE> --reuse-values --set configuration.provider=<NEW PROVIDER>
 ```
 
 #### Using Helm 2
@@ -107,6 +107,8 @@ helm install vmware-tanzu/velero \
 --set initContainers[0].volumeMounts[0].mountPath=/target \
 --set initContainers[0].volumeMounts[0].name=plugins 
 ```
+
+Users of zsh might need to put quotes around key/value pairs.
 
 ##### Option 2) YAML file
 
--- a/charts/velero/README.md
+++ b/charts/velero/README.md
@@ -24,6 +24,9 @@ The default configuration values for this chart are listed in values.yaml.
 
 See Velero's full [official documentation](https://velero.io/docs/v1.5/basic-install/). More specifically, find your provider in the Velero list of [supported providers](https://velero.io/docs/v1.5/supported-providers/) for specific configuration information and examples.
 
+#### Set up Helm
+
+See the main [README.md](https://github.com/vmware-tanzu/helm-charts#kubernetes-helm-charts-for-vmware-tanzu).
 
 #### Using Helm 3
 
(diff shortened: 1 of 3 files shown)


charts/velero/README.md
```diff
@@ -50,9 +53,10 @@ helm install vmware-tanzu/velero \
 --set initContainers[0].image=velero/velero-plugin-for-<PROVIDER NAME>:<PROVIDER PLUGIN TAG> \
 --set initContainers[0].volumeMounts[0].mountPath=/target \
 --set initContainers[0].volumeMounts[0].name=plugins \
---generate-name
 ```
 
+Users of zsh might need to put quotes around key/value pairs.
```
jenting, line 58: I never seen users had reported this issue before.
Besides, could we make sure it always need to add quotes? So we could remove the "might" term.

codegold79: When I use zsh, I didn't need quotes for every key/value pair. I only needed quotes for the ones that indexed an array. For example, these work:

```
[...clipped...]
 --set configuration.volumeSnapshotLocation.config.region=us-east-2 \
 --set "initContainers[0].name=velero-plugin-for-aws" \
[...clipped...]
```

Bash requires no quotes at all.

jenting: I use fish shell and did not need any quotes.
Thanks for your explanation.

jenting: BTW, which helm version you are using?

codegold79: There are a lot of MacBook users who use zsh because it is the default shell for OSX Catalina and newer. I wonder if I'm the only zsh user with the quotes problem.

Here are the versions I'm using:
```
❯ helm version
version.BuildInfo{Version:"v3.5.1", GitCommit:"32c22239423b3b4ba6706d450bd044baffdcf9e6", GitTreeState:"dirty", GoVersion:"go1.15.7"}
❯ zsh --version
zsh 5.7.1 (x86_64-apple-darwin19.0)
```

When I don't use quotes, I get this error:
```
zsh: no matches found: initContainers[0].name=velero-plugin-for-aws
```

jenting: You're right, it's `zsh` behavior https://stackoverflow.com/a/65687265/4656380.

charts/velero/README.md
```diff
@@ -65,7 +69,7 @@ helm install vmware-tanzu/velero --namespace <YOUR NAMESPACE> -f values.yaml --g
 If a value needs to be added or changed, you may do so with the `upgrade` command. An example:
 
 ```bash
-helm upgrade vmware-tanzu/velero <RELEASE NAME> --namespace <YOUR NAMESPACE> --reuse-values --set configuration.provider=<NEW PROVIDER>
+    helm upgrade vmware-tanzu/velero <RELEASE NAME> --namespace <YOUR NAMESPACE> --reuse-values --set configuration.provider=<NEW PROVIDER>
```
jenting, line 72: This change would break the alignment.
image

charts/velero/README.md
jenting: I'd prefer to adding this to note _or_ add it under both helm2 and helm3.

jenting (changes requested):

charts/velero/README.md
codegold79: I'll be sure to have a note in both the Helm 2 and Helm 3 sections.

charts/velero/README.md
codegold79: Good catch. I'll revert the indent.

jenting (approved):
lgtm

jenting:
@cpanato @carlisia Please take a look

carlisia (approved):
Thanks for the change and also improving the docs @codegold79! 👍

PR #221, opened 2021-02-23
[velero] Add container level SecurityContext

mkilchhofer:
This PR adds the ability for confining the SecurityContext of the velero pod even more. Some settings (like read-only filesystem and dropping all linux caps) need to be done on container level. 

It seems that the binary `/velero` wants to write to `/tmp` when loading the plugins (eg. `velero-plugin-for-aws`):

```log
level=error msg="plugin init error" cmd=/velero error="open /tmp/plugin431659487: read-only file system" logSource="pkg/plugin/clientmgmt/logrus_adapter.go:89" pluginName=velero
An error occurred: Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
```

Because of this, I mount an `emptyDir` volume once the read-only filesystem option is activated.

#### Checklist
- [X] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[velero]`)


jenting (changes requested):
Would you mind to bump the chart version? Thank you.

jenting (dismissed):

carlisia (changes requested):
@mkilchhofer thank you for this.

Couple requests: please add the changes to the restic deployment as well, and pls bump the chart, thanks.

mkilchhofer:
> @mkilchhofer thank you for this.
> 
> Couple requests: please add the changes to the restic deployment as well, and pls bump the chart, thanks.

I'd love to do that, but it is a bit tricky. The values `securityContext` (`.Values.securityContext` and `.Values.restic.securityContext`) are used in a different context:
- `.Values.securityContext`: affects pod-level SecurityContext of the velero pods
- `.Values.restic.securityContext`: affects container-level SecurityContext of the restic pods

-> I'd propose to use `podSecurityContext` and `containerSecurityContext` as your vmware colleagues from Bitnami use. If you are okay with that, shall I implement it in a backward-compatible way or do we simlply release a new major?

carlisia:
@jenting / @cpanato I'd appreciate your 👀  on this one.

jenting:
> I'd love to do that, but it is a bit tricky. The values `securityContext` (`.Values.securityContext` and `.Values.restic.securityContext`) are used in a different context:
> 
> * `.Values.securityContext`: affects pod-level SecurityContext of the velero pods
> * `.Values.restic.securityContext`: affects container-level SecurityContext of the restic pods

I need to take extra time to investigate what's the difference between pod-level and container-level SecurityContext.
But, thanks for your contribution for improve security.

> -> I updated the PR and proposed to use `podSecurityContext` and `containerSecurityContext` as your vmware colleagues from Bitnami use. I implement it in a backward-compatible way. We also could remove the compat code and bump the major version. WDYT?

Grant-Jia:
Hi @jenting Any progress on the review of this PR. We need this feature and add SecurityContext on container level. Thanks.

charts/velero/values.yaml
```diff
@@ -327,9 +336,14 @@ restic:
   # SecurityContext to use for the Velero deployment. Optional.
   # Set fsGroup for `AWS IAM Roles for Service Accounts`
   # see more informations at: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
-  securityContext: {}
+  podSecurityContext:
+    runAsUser: 0
```
jenting, line 340: Any reason why change it to `runAsUser: 0` (root user)?
From the Velero container image, by default, it runs as `nobody`.
https://github.com/vmware-tanzu/velero/blob/v1.6.0/Dockerfile#L60

charts/velero/values.yaml
```diff
@@ -61,9 +61,18 @@ initContainers: []
...
+podSecurityContext: {}
   # fsGroup: 1337
 
+# Container Level Security Context for the 'velero' container of the Velero deployment. Optional.
+# See: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
+containerSecurityContext: {}
+  # allowPrivilegeEscalation: false
+  # capabilities:
+  #   drop: ["ALL"]
+  #   add: []
```
jenting, line 73: I'm not sure should we list these values in the values.yaml?
I believe it does not include the full security context.
If not, probably we could consider removing it. WDYT?

charts/velero/values.yaml
mkilchhofer: This is the `restic:` section of the values file. Restic did already run as root (and I think this is necessary to backup things), but I moved it from beeing hardcoded (see [restic-daemonset.yaml@Line47](https://github.com/vmware-tanzu/helm-charts/pull/221/files#diff-098ff3169e23f0fd8394cfaab72d002c19232c7f59d7e30321df2162df86aa81L47)) to the `values.yaml` where users are able to change this.

charts/velero/values.yaml
mkilchhofer: These are settings most users want to set for security reasons. But I did not want to change defaults to not break any edge-case setup.

jenting:
LGTM, thanks for the effort.

jenting (dismissed):

jenting:
@carlisia @zubron  Please take a look if you're available :pray:

mkilchhofer:
Hi all, PR was approved by @jenting and I now updated the base branch again. Can we merge this? :-)

jenting (approved):

zubron (approved):
Thanks, @mkilchhofer!

PR #246, opened 2021-04-14
[velero] Fix `helm upgrade` does not upgrade the CRDs

### LINKED ISSUE #197: `helm upgrade` does not upgrade the CRDs
**What steps did you take and what happened:**

When performs `helm upgrade`, the Velero CRDs do not upgrade.

```
helm install -f values.yaml --version 2.12.2
kubectl get crds backupstoragelocations.velero.io -oyaml > backupstoragelocations.velero.io.2.12.2.yaml

helm upgrade -f values.yaml --version 2.14.4
kubectl get crds backupstoragelocations.velero.io -oyaml > backupstoragelocations.velero.io.2.14.4.yaml

diff backupstoragelocations.velero.io.2.12.2.yaml backupstoragelocations.velero.io.2.14.4.yaml
```

**What did you expect to happen:**

The CRDs be upgraded when performing `helm upgrade`.

**The output of the following commands will help us better understand what's going on**:

N/A

N/A

**Environment:**

- helm version (use `helm version`): 
- helm chart version and app version:
  - helm2 (use `helm list`):
  - helm3 (use `helm list -n <YOUR NAMESPACE>`):
- Kubernetes version (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

jenting:

Run a _Job_ to apply the CRDs against the Velero container image the user specify.
- On `helm install`, the user might _change_ the velero container image tag, we should update the CRDs to make sure they match against the Velero server.
-  On `helm upgrade`, the velero container image might _upgrade_ to a newer version, we should update the CRDs to make sure they match against the Velero server.
- On `helm rollback`, the velero container image might _downgrade_ to an old version, we should update the CRDs to make sure they match against the Velero server.

Then, we configured these CRs to the cluster (after the applying the CRDs finished):
1. backupstoragelocations.yaml
1. schedules.yaml
1. volumesnapshotlocation.yaml

We _must_ make sure the install/upgrade/rollback CRDs job finished then apply these CRs manifest.
According to https://helm.sh/docs/topics/charts_hooks/, we add the `post-install`, `post-upgrade`, and `post-rollback` to these CRs to make sure that the CRs will wait for the install/upgrade/rollback CRDs job runs to completion and then, apply the CRs manifest.

BTW, since velero CLI installed CRDs are with labels `component: velero`. To make sure both CLI installed Velero _or_ helm installed Velero having the same labels, we add label `component: velero` to the CRDs.

fixes #197

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/templates/backupstoragelocation.yaml
+++ b/charts/velero/templates/backupstoragelocation.yaml
@@ -4,7 +4,7 @@ kind: BackupStorageLocation
 metadata:
   name: {{ include "velero.backupStorageLocation.name" . }}
   annotations:
-    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook": post-install,post-upgrade,post-rollback
     "helm.sh/hook-delete-policy": "before-hook-creation"
   labels:
     app.kubernetes.io/name: {{ include "velero.name" . }}
--- a/charts/velero/templates/backupstoragelocation.yaml
+++ b/charts/velero/templates/backupstoragelocation.yaml
@@ -5,7 +5,7 @@ metadata:
   name: {{ include "velero.backupStorageLocation.name" . }}
   annotations:
     "helm.sh/hook": post-install,post-upgrade,post-rollback
-    "helm.sh/hook-delete-policy": "before-hook-creation"
+    "helm.sh/hook-delete-policy": before-hook-creation
   labels:
     app.kubernetes.io/name: {{ include "velero.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
--- a/charts/velero/templates/upgrade-crds.yaml
+++ b/charts/velero/templates/upgrade-crds.yaml
@@ -1,0 +1,32 @@
+{{- if .Release.IsUpgrade }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "velero.fullname" . }}-upgrade-crds
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "velero.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ include "velero.chart" . }}
+spec:
+  template:
+    metadata:
+      name: velero-upgrade-crds
+    spec:
+      serviceAccountName: {{ include "velero.serverServiceAccount" . }}
+      containers:
+        - name: velero
+      {{- if .Values.image.digest }}
+          image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+      {{- else }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+      {{- end }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - /velero
+          args:
+            - install
+            - --crds-only
+      restartPolicy: OnFailure
+{{- end }}
--- a/charts/velero/templates/upgrade-crds.yaml
+++ b/charts/velero/templates/upgrade-crds.yaml
@@ -15,7 +15,7 @@ spec:
       name: velero-upgrade-crds
     spec:
       serviceAccountName: {{ include "velero.serverServiceAccount" . }}
-      containers:
+      initContainers:
         - name: velero
       {{- if .Values.image.digest }}
           image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
--- a/charts/velero/templates/upgrade-crds.yaml
+++ b/charts/velero/templates/upgrade-crds.yaml
@@ -24,9 +24,25 @@ spec:
       {{- end }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
-            - /velero
-          args:
-            - install
-            - --crds-only
+            - /bin/sh
+            - -c
+            - /velero install --crds-only --dry-run -o yaml > /tmp/crds.yaml
+          volumeMounts:
+            - mountPath: /tmp
+              name: crds
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
+          imagePullPolicy: IfNotPresent
+          command:
+            - /bin/sh
+            - -c
+            - kubectl apply -f /tmp/crds.yaml
+          volumeMounts:
+            - mountPath: /tmp
+              name: crds
+      volumes:
+        - name: crds
+          emptyDir: {}
       restartPolicy: OnFailure
 {{- end }}
--- a/charts/velero/templates/upgrade-crds.yaml
+++ b/charts/velero/templates/upgrade-crds.yaml
@@ -1,9 +1,11 @@
-{{- if .Release.IsUpgrade }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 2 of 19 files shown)


carlisia:
@jenting  did you have a chance to test this?

carlisia (dismissed):
👍

cpanato (dismissed):

charts/velero/templates/upgrade-crds.yaml
```diff
@@ -0,0 +1,48 @@
...
+          command:
+            - /bin/sh
+            - -c
+            - /velero install --crds-only --dry-run -o yaml > /tmp/crds.yaml
+          volumeMounts:
+            - mountPath: /tmp
+              name: crds
+      containers:
+        - name: kubectl
+          image: docker.io/bitnami/kubectl:1.14.1
```
zubron, line 35: Is there a reason for using this specific version of the `kubectl` image? I'm not totally clear on the version support matrix for kubectl and cluster versions. If possible, it would be preferable to use a newer image, or at least the latest patch release of this one. My concern is that the 1.14.* series was last updated a year ago and could contain security vulnerabilities. 1.16-1.18 have been [deprecated upstream](https://github.com/bitnami/bitnami-docker-kubectl#branch-deprecation-notice) but Bitnami are still making builds of those images for a short time which will include recent security vulnerability patches.

charts/velero/values.yaml
```diff
@@ -119,7 +119,7 @@ configuration:
   provider:
 
   # Parameters for the `default` BackupStorageLocation. See
-  # https://velero.io/docs/v1.5/api-types/backupstoragelocation/
+  # https://velero.io/docs/main/api-types/backupstoragelocation/
```
zubron, line 122: Can this link, and the one below, be changed to point to the 1.6 branch? I wouldn't want to direct people to main incase there are incompatible changes.

charts/velero/templates/upgrade-crds.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - /bin/sh
+            - -c
+            - kubectl apply -f /tmp/crds.yaml
+          volumeMounts:
+            - mountPath: /tmp
+              name: crds
+      volumes:
+        - name: crds
+          emptyDir: {}
+      restartPolicy: OnFailure
```
zubron, line 47: This job should also set a [`backoffLimit`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) so that it doesn't keep retrying in the case where retrying won't fix it (e.g. the API server rejects the new CRDs).

charts/velero/templates/upgrade-crds.yaml
```diff
@@ -0,0 +1,48 @@
+{{- if .Release.IsUpgrade }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "velero.fullname" . }}-upgrade-crds
```
zubron, line 5: Will this create a fixed job name with no unique components? My concern is that there is no TTL specified on this job which means that it will persist after completion. If the user attempts another upgrade, this will fail as the job name already exists.

charts/velero/templates/backupstoragelocation.yaml
```diff
@@ -4,7 +4,7 @@ kind: BackupStorageLocation
 metadata:
   name: {{ include "velero.backupStorageLocation.name" . }}
   annotations:
-    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook": post-install,post-upgrade,post-rollback
```
zubron, line 7: Are the `helm.sh/hook` changes here and in other files relevant to this change? I'm not super familiar with helm lifecycle management so I don't know if there's something I'm missing here :)

zubron (changes requested):
Thanks, @jenting! This is looking good so far. I just have a couple of questions and some requested changes.

charts/velero/templates/upgrade-crds.yaml
jenting: Ref to https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies
By default, the `helm.sh/hook-delete-policy` is `before-hook-creation`, which means the previous upgrade Job would be deleted before this new Job is launched.

I could verify it again.

charts/velero/templates/upgrade-crds.yaml
jenting: Because I choose the same kubectl image as https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/templates/cleanup-crd.yaml#L26.

Yes,  I could change to a newer image.

charts/velero/templates/upgrade-crds.yaml
zubron: Ahh, sorry, I wasn't aware of that. If helm already manages this then that's fine :) thanks for the explanation!

charts/velero/templates/backupstoragelocation.yaml
jenting: Since helm support `upgrade` and `rollback`.
I thought that the user is possible to roll back to the previous version once the user found any issue in the newer version.
Therefore, the CRDS should be rollback to the previous one also.

charts/velero/templates/upgrade-crds.yaml
jenting: I don't have a strong opinion on which kubectl version to use since Velero supports the Kubernetes version from v1.12+.

zubron: If we're using that image elsewhere, then maybe we can continue using it for now. There might be some risk in upgrading, so we can stick with this for now (but bump to 1.14.3 to incorporate new fixes). We can address upgrading both this usage and the usage in the cleanup template at a later time :+1:

charts/velero/templates/upgrade-crds.yaml
jenting: Thank you, it's really important.

charts/velero/templates/upgrade-crds.yaml
zubron: I noticed you just upgraded 1.16, if you've tried it out and it works, that's also fine :smile:

charts/velero/templates/backupstoragelocation.yaml
zubron: Thanks, that makes sense!

charts/velero/templates/upgrade-crds.yaml
zubron: Thanks for adding that!

zubron (approved):
Thanks for making the changes, @jenting! And thanks for teaching me new things about helm :smile:

carlisia:
@jenting I'll wait for this to change to ready to review to take a look.

charts/velero/templates/upgrade-crds.yaml
jenting: I tried v1.14.3 on Kubernetes 1.20.5 cluster, it works well.

jenting:
I thought of a scenario that during fresh install `helm install`, the user might change the default container image tag (not the latest one). In this scenario, we should update the CRDs by a _Job_ also.

jenting:
@carlisia @zubron this is ready to review, I have verified it on lots of scenarios I could think of.

zubron (approved):
Thanks for your efforts on this, @jenting!

carlisia (approved):
Thanks for helping us crack this nut!

PR #257, opened 2021-04-22
[velero] Allow better customization for upgrade-crds job

uncycler:
This PR allows the user to customize the upgrade-crds job with labels, annotations and kubectl image.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [X] Chart Version bumped
- [X] Variables are documented in the values.yaml or README.md
- [X] Title of the PR starts with chart name (e.g. `[velero]`)

PR #259, opened 2021-04-29
[velero] Make upgradeCrds an optional hook

nirnanaaa:
The upgrade CRD hook is not always necessary (especially when using helms' `skip-crds` config option and deploying the crds manually).

In addition to that the upgrade hook was pretty limited in its configuration, so i decided to also
- add support for `imagePullSecrets` (which COULD be obtained through the service account, but since the deployment also supports it, I decided to add it to the job as well)
- add the ability to override the kubectl image (especially necessary in on-prem environments or if you want to use a newer kubectl version, closer to the version running on your cluster)

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.6.0
 description: A Helm chart for velero
 name: velero
-version: 2.19.2
+version: 2.19.3
 home: https://github.com/vmware-tanzu/velero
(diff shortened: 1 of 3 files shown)


nirnanaaa:
sorry, didn't see #257

jenting:
No worries, I'm thinking is there any variable in helm that the chart known `--skip-crds` flag passing.
So, we could ignore not create the upgrade CRDs Job by helm native way without the need to define a new variable in values.yaml.

PR #266, opened 2021-05-17
[velero] Update auto assign of reviewers

carlisia:
Signed-off-by: Carlisia <carlisia@grokkingtech.io>

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/.github/auto-assignees.yml
+++ b/.github/auto-assignees.yml
@@ -25,4 +25,4 @@ options:
     - wip
     - DO NOT MERGE
   enable_group_assignment: true
-  number_of_reviewers: 3
+  number_of_reviewers: 2
--- a/.github/auto-assignees.yml
+++ b/.github/auto-assignees.yml
@@ -1,0 +1,28 @@
+---
+# This assigns a PR to its author
+addAssignees: author
+
+reviewers:
+  # The default reviewers
+  defaults:
+    - maintainers
+
+  groups:
+    maintainers:
+      - carlisia
+      - jenting
+
+    other-maintainers:
+      - ashish-amarnath
+      - zubron
+      - dsu-igeek
+      - sseago
+
+options:
+  ignore_draft: true
+  ignored_keywords:
+    - WIP
+    - wip
+    - DO NOT MERGE
+  enable_group_assignment: true
+  number_of_reviewers: 3
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.6.0
 description: A Helm chart for velero
 name: velero
-version: 2.21.0
+version: 2.21.1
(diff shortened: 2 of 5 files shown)


.github/auto-assignees.yml
```diff
@@ -0,0 +1,28 @@
...
+      - sseago
+
+options:
+  ignore_draft: true
+  ignored_keywords:
+    - WIP
+    - wip
+    - DO NOT MERGE
+  enable_group_assignment: true
+  number_of_reviewers: 3
```
zubron, line 28: Do we still need 3 reviewers here? We're using 2 for the main velero project and the new auto-request-review action seems to handle that well (it looks like there was an issue with the previous action being used).

.github/auto-assignees.yml
```diff
@@ -0,0 +1,28 @@
...
+  # The default reviewers
+  defaults:
+    - maintainers
+
+  groups:
+    maintainers:
+      - carlisia
+      - jenting
+
+    other-maintainers:
```
zubron, line 15: If we set the number of reviewers to 2, would anyone from this group be selected to review?

carlisia: You're on the right track. No, it would never overflow into the "other-maintainers" because the default would satisfy the condition.

.github/auto-assignees.yml
carlisia: Because this repo has the potential to be de-prioritized, I thought it would be best to require the reviews of 3 people.  I should've documented this in the PR. 

The 2nd reviewer who gets to it first can merge the PR. This is not requiring 3, only requesting 3. We still only require 2.

WDYT?

jenting: I'd prefer using 2.
We could use Slack _or_ GitHub notification to ask other maintainers to review.

carlisia: Ok I'll change it to 2.

.github/auto-assignees.yml
carlisia: Actually, reading the instructions again, I *think* if we set to 2 then it will randomly pick from either group. It would be ideal if it could pick 1 from each group so that way we could always have at least 1 of the more regular helm maintainers in the rotation. We can try this and see if it works this way or not.

charts/velero/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.6.0
 description: A Helm chart for velero
 name: velero
-version: 2.21.0
+version: 2.21.1
```
jenting, line 5: I don't think we need to bump the Chart version since it's a CI enhancement.

carlisia: I was trying to remember if we had to or not. Will revert.

jenting (approved):

zubron (approved):

PR #305, opened 2021-09-18
[velero] Allow to define resource requests/limits for the containers in the upgrade/cleanup job

weseven:
This PR allows to define resource requests and limits for the containers in the upgrade and cleanup jobs.
This is necessary in kubernetes clusters where an admission controller prevents the creation of pods without proper resource requests and limits.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/velero/templates/cleanup-crds.yaml
+++ b/charts/velero/templates/cleanup-crds.yaml
@@ -53,6 +53,10 @@ spec:
               kubectl delete volumesnapshotlocation --all;
               kubectl delete podvolumerestore --all;
               kubectl delete crd -l app.kubernetes.io/name=velero;
+          {{- with .Values.kubectl.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
       restartPolicy: OnFailure
       {{- with .Values.nodeSelector }}
       nodeSelector:
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.6.3
 description: A Helm chart for velero
 name: velero
-version: 2.23.8
+version: 2.23.9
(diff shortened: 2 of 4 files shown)


charts/velero/templates/cleanup-crds.yaml
```diff
@@ -53,6 +53,10 @@ spec:
               kubectl delete volumesnapshotlocation --all;
               kubectl delete podvolumerestore --all;
               kubectl delete crd -l app.kubernetes.io/name=velero;
+          {{- with .Values.kubectl.resources }}
```
sseago, line 56: Should these additions be put behind a conditional so that empty "resources" fields aren't added? I'm not 100% sure that it matters, but that way there's no change in yaml for cases where the optional resources aren't specified:
```
    {{- if .Values.kubectl.resources }}
```

weseven: Hi,
 thanks for the review.

I had the same concern but testing with helm template if resources is empty (either `resources: {}` like in values.yaml or simply `resources:`) nothing gets included in the final manifest. Basically, if there's something defined it gets converted in yaml and indented with 12 spaces; if there's nothing, nothing gets added as expected. 

The same style of inclusion is used freely in the chart ( e.g. for the resources of the main deployment here https://github.com/vmware-tanzu/helm-charts/blob/0c49944a8ee8ae9ac92a508e0344a2249d05fafa/charts/velero/templates/deployment.yaml#L112, the same goes for nodeSelector,affinity, tolerations, etc.).

sseago (approved):

charts/velero/templates/cleanup-crds.yaml
jenting: I prefer using `with` syntax personally.

jenting (approved):
LGTM
Thanks for the contribution @weseven

PR #313, opened 2021-10-06
[velero] Follow istio naming convention

sathieu:
See https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

#### Checklist

- [x] Chart Version bumped


dsu-igeek:
Looks like if you use the regular installer (velero install) this port is named "metrics" - I'd like to have everything be the same and I think we should discuss a little about whether this will break anything else (since Helm and the CLI installer are doing things differently, hopefully no one is dependent on the port name).    Any thoughts on if linkerd will work with this as well?

PR #323, opened 2021-11-23
[velero] allow setting initContainers as a template

tareqhs:

A container is a complex structure that itself is usually templated with numerous references to values and other subtemplates. In the values of velero chart `initContainers` is given as a list which means it can't be overridden partially - the whole definition of the init containers must be copied every time a tiny override is needed.

This change allows setting the value of `initContainers` as a string which in turn causes the chart to render it as a template, e.g.:
```yaml
initContainers: |-
  - name: myplugin
    image: {{ .Values.myplugin.image.repository }}:{{ .Values.myplugin.image.tag }}
    resources:
      {{- .Values.myplugin.resources | toYaml | nindent 6 }}
    ...
myplugin:
  image:
    ...
  resources:
    ...
```
and then the init container itself can be customized conveniently by regular values selectively.

Note: I've pumped the minor version and not the patch version because this introduces a change in the schema of values.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.7.0
 description: A Helm chart for velero
 name: velero
-version: 2.26.5
+version: 2.27.0
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/templates/deployment.yaml
+++ b/charts/velero/templates/deployment.yaml
@@ -184,7 +184,11 @@ spec:
       dnsPolicy: {{ .Values.dnsPolicy }}
 {{- if .Values.initContainers }}
       initContainers:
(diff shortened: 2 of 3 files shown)


jenting:
Thank you @tareqhs 

Could you please some examples by using `initContainers` as a string here?
After that, we could have a reference for it once there are other users who want to use this.
Appreciate.

tareqhs:
hi @jenting , the example is given in the PR description. do you want me to document it somewhere else as well?

jenting:
> ```yaml
> |-
> ```

I see, so the main key is the `|-`, right?

jenting (dismissed):
LGTM, thank you @tareqhs

tareqhs:
thank you @jenting . @reasonerjt, @zubron - any feedback?

jenting (approved):

sseago (approved):

HaveFun83:
My humble opinion. 
Normally you don't specify any custom values `.Values.myplugin` within an upstream chart.
Instead you create new chart skeleton with your own values and write a dependency for velero into your chart.yaml.
Then you could use the helm templating within your values from the upsteam chart without this hacky function.

PR #370, opened 2022-05-10
[velero] Only install ServiceMonitor when monitoring API is available

ariep:
Signed-off-by: Arie Peterson <arie@greenhost.nl>

With this change, the ServiceMonitor is only installed if the corresponding monitoring API is available on the cluster (and the proper helm values are enabled, as before). This way, we can safely enable the servicemonitor in our helm values for velero that are used on many clusters, some of which have the monitoring API/prometheus and some of which do not. The unattractive alternative would be to have some complex mechanism to conditionally set the helm value depending on the cluster configuration.

This is inspired by the [same construct in another helm chart](https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/service-admin.yaml#L28).

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.8.1
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 2.29.4
+version: 2.29.5
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.8.1
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 2.29.5
+version: 2.29.6
 home: https://github.com/vmware-tanzu/velero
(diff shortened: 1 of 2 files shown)


sseago (approved):

jenting (commented):
Code LGTM, please help bump the chart version.

ariep:
> Code LGTM, please help bump the chart version.

Have to run to keep up :). Done!

jenting (approved):

jenting (changes requested):
CI failed, could you check?

ariep:
> CI failed, could you check?

I'm not really sure what's going wrong. From one of the CI jobs:
```
==> Linting charts/velero
Error:  templates/: parse error at (velero/templates/servicemonitor.yaml:1): unclosed action
```
However, when I run `helm lint` or `helm lint --strict` locally (with helm 3.8.2), linting succeeds. I found [one issue](https://github.com/arttor/helmify/issues/11) with the same error message, which was solved by upgrading helm. Also, I notice that the CI job uses `quay.io/helmpack/chart-testing:v3.0.0`, which is almost two years old. Would you be willing to update that, so it uses a newer helm version for the chart testing internally, and retry the CI with that?

jenting:
Sure, we could update the helm version.

jenting:
I create #373 to update the helm version.

jenting:
@ariep Please rebase to the main branch.

jenting (approved):

ywk253100 (approved):

jenting:
Oops, we forget to bump the chart version, so the chart release failed.
We need to check on why CI did not catch this error.

PR #382, opened 2022-06-15
velero: ensure that the plugin is installed

omichels:
#### Checklist
- [x] Chart Version bumped


jenting:
Thanks to your contribution, however, it means we need to handle the plug-ins version as well.

If you were installing with the correct command, the issue isn't exist, correct?

jenting:
Correct me if I'm wrong, but I still think this PR doesn't need it.
If you were installing with the correct command, the issue doesn't exist, correct?

PR #408, opened 2022-10-08
[velero] Add Termination Grace Period Helm Value

celestialorb:
Adds in the ability to specify the termination grace period to the Helm chart. Primary use case is to allow Velero to finish a backup before terminating in case it takes more than the default, previously immutable, 30 seconds to perform a backup.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -86,6 +86,9 @@ lifecycle: {}
 # Pod priority class name to use for the Velero deployment. Optional.
 priorityClassName: ""
 
+# The number of seconds to allow for graceful termination of the pod. Optional.
+terminationGracePeriodSeconds: 3600
+
 # Tolerations to use for the Velero deployment. Optional.
 tolerations: []
 
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.9.2
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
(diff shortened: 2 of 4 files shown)


jenting:
Could you please add the restic daemonset as well? Thank you 😊

celestialorb:
> Could you please add the restic daemonset as well? Thank you blush

Done!

charts/velero/values.yaml
```diff
@@ -86,6 +86,9 @@ lifecycle: {}
 # Pod priority class name to use for the Velero deployment. Optional.
 priorityClassName: ""
 
+# The number of seconds to allow for graceful termination of the pod. Optional.
+terminationGracePeriodSeconds: 3600
```
jenting, line 90: Since it's an optional field and the Kubernetes `terminationGracePeriodSeconds` default value is 30 seconds. I think 3600 seconds changes the default time.

celestialorb: True, it does, but shouldn't the default of the Helm chart be suited to the use case? Under common conditions I think Velero can easily take over thirty seconds to create a backup.

Or are you suggesting that this needs to be a minor version bump as opposed to a patch version bump?

jenting: > True, it does, but shouldn't the default of the Helm chart be suited to the use case? Under common conditions I think Velero can easily take over thirty seconds to create a backup.

Agree with you.

jenting (approved):
Thank you @celestialorb

ywk253100 (approved):

PR #411, opened 2022-10-11
[velero] Adding garbageCollectionFrequency to velero server

arthurburle:
Signed-off-by: Arthur Burle <arthurburle@gmail.com>

The purpose of this PR is to allow the Chart user to set the flag `garbageCollectionFrequency` to the `velero server` command.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -304,8 +304,6 @@ configuration:
   storeValidationFrequency:
   # `velero server` default: 1h
   garbageCollectionFrequency:
-  # `velero server` default: []
-  extraArgs: []
   #
 
   # additional key/value pairs to be used as environment variables such as "AWS_CLUSTER_NAME: 'yourcluster.domain.tld'"
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -302,6 +302,10 @@ configuration:
   disableControllers:
   # `velero server` default: 1m
   storeValidationFrequency:
+  # `velero server` default: 1h
+  garbageCollectionFrequency:
+  # `velero server` default: []
+  extraArgs: []
   #
 
   # additional key/value pairs to be used as environment variables such as "AWS_CLUSTER_NAME: 'yourcluster.domain.tld'"
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.9.2
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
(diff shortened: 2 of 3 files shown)


charts/velero/values.yaml
```diff
@@ -302,6 +302,10 @@ configuration:
   disableControllers:
   # `velero server` default: 1m
   storeValidationFrequency:
+  # `velero server` default: 1h
+  garbageCollectionFrequency:
+  # `velero server` default: []
+  extraArgs: []
```
jenting, line 308: Providing extraArgs is brilliant.
But I am thinking that should we always keep helm chart args aligned with current velero server-supported arguments?

Also, if we want to support `extraArgs`, add `garbageCollectionFrequency` reasonable?

jenting (changes requested):
Thank you for your contribution.
You need to bump the chart version.

charts/velero/values.yaml
arthurburle: Hi @jenting !
I'd say it depends on the architecture. Since you guys were following that pattern I decided to keep that way while still adding the extraArgs as a backup plan to prevent other missing values in the future.

If you think it's better that way I can remove the garbageCollectionFrequency part. It also sounds good for me.

arthurburle:
> Thank you for your contribution. You need to bump the chart version.

Hey @jenting, please let me know if I bumped the version correctly and also if there is anything else that I need to do.
Also, thanks for answering this PR so quickly.

charts/velero/values.yaml
jenting: 🤔 I don't have a strong opinion on this. But I'd prefer to use garbageCollectionFrequency part rather than extraArgs.

WDYT? @ywk253100 @reasonerjt

reasonerjt: I would vote for the `garbageCollectionFrequency` approach, thanks @jenting .

arthurburle: If you guys wanna check another repo that uses this approach you can check [cert-manager](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L109).
But any case you guys decide will address my issue. Let me know what you decide so I can make the necessary changes tomorrow. (3am here)

arthurburle: Hey @jenting, I just corrected the PR to follow the pattern you guys prefered.
Please tell me if there's anything else I should do.

jenting (approved):
Thank you @arthurburle

reasonerjt (approved):
lgtm

PR #414, opened 2022-10-27
[velero] Bump velero version v1.10.0

qiuming-best:
Signed-off-by: Ming <mqiu@vmware.com>

#### Checklist
- [x] Chart Version bumped


qiuming-best:
I'll keep it in draft until v1.10.0 is been released

charts/velero/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: 1.9.2
+appVersion: 1.10.0
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 2.32.1
+version: 2.32.2
```
jenting, line 6: Since the Velero bumps the minor version, the helm chart should bump the minor version.

charts/velero/values.yaml
```diff
@@ -284,10 +284,12 @@ configuration:
   # here if using a non-default value. The `velero server` default values are shown in the
   # comments below.
   # --------------------
+  # `velero server` default: restic
+  uploaderType:
   # `velero server` default: 1m
   backupSyncPeriod:
   # `velero server` default: 1h
-  resticTimeout:
+  fsBackupTimeout:
```
jenting, line 292: It will break the backward-compatible if the user does not update their values.yaml before upgrading to this new helm chart. 🤔 [1]

charts/velero/values.yaml
```diff
@@ -319,11 +321,11 @@ configuration:
   # Set log-format for Velero pod. Default: text. Other option: json.
   logFormat:
 
-  # Set true for backup all pod volumes without having to apply annotation on the pod when used restic Default: false. Other option: false.
-  defaultVolumesToRestic:
+  # Set true for backup all pod volumes without having to apply annotation on the pod when used file system backup Default: false. Other option: false.
+  defaultVolumesToFsBackup:
```
jenting, line 325: [1]

charts/velero/values.yaml
```diff
@@ -319,11 +321,11 @@ configuration:
...
 
-  # How often 'restic prune' is run for restic repositories by default. Default: 168h. Optional.
-  defaultResticPruneFrequency:
+  # How often repository maintain is run for repositories by default.
+  defaultRepoMaintainFrequency:
```
jenting, line 328: [1]

charts/velero/crds/restores.yaml
```diff
@@ -1,3 +1,5 @@
+
```
jenting, line 1: What if the existing cluster contains the restores CRs? Can helm remove the restores CRDs if there are restored CRs within the cluster?

charts/velero/crds/backuprepositories.yaml
```diff
@@ -8,20 +8,23 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
   creationTimestamp: null
-  name: resticrepositories.velero.io
+  name: backuprepositories.velero.io
```
jenting, line 11: What if the existing cluster contains the resticrepositories CRs? Can helm remove the resticrepositories CRDs if there are resticrepositories CRs within the cluster?

jenting (changes requested):
I believe the user would encounter many upgrade issues if there were many existing Velero CRs within the cluster.

charts/velero/Chart.yaml
qiuming-best: Thanks, I'll modify it

charts/velero/crds/restores.yaml
qiuming-best: We wouldn't remove existing CRs when upgrading to a higher version, is it better to keep the restores CRDs instead of removing them from the cluster for the users? users could still get restores by velero cli.

charts/velero/crds/backuprepositories.yaml
qiuming-best: We wouldn't remove existing CRs when upgrading to a higher version, but user could not using `velero restic repo get` to list the resticrepositories.

charts/velero/values.yaml
qiuming-best: I don't know what is the best practice for backward-compatible if the name of fields in values.yaml has changed?

Maybe is easier to fail the helm upgrade when using legacy names in values.yaml? or I should add lots of conditional judgment in `templates` directories when set with legacy names? that would be lots of work?

jenting: It'd be nice to auto-translate the legacy names into new names.

@qiuming-best How is the vmware-tanzu/velero handling this case? We could try to make an alignment with the same approach.

charts/velero/crds/backuprepositories.yaml
jenting: Since the CRD changes from `resticrepositories.velero.io` to `backuprepositories.velero.io`.
If I remember correctly, helm upgrade will
- install a new CRD `backuprepositories.velero.io`
- remove the CRD `resticrepositories.velero.io`

Therefore, is there were exiting resticrepositories CRs within the cluster, I _guess_ the helm upgrade would fail because the CRD `resticrepostiories.velero.io` can't be removed successfully (because there are resticrepositories CRs existed).

charts/velero/crds/restores.yaml
jenting: [2]

charts/velero/crds/backuprepositories.yaml
qiuming-best: I've tested that it will still keep resticrepositories.velero.io CRD and CRs after successfully upgrading, also backuprepositories.velero.io could be created

charts/velero/values.yaml
qiuming-best: As we've discussed, we add verification names of variables in NOTES.txt.
if a user is still using the legacy names it will install or upgrade fail with an error message in detail on invalid names of variables.
Compared with translating legacy names this will make the work more accessible, and from v1.10, Velero could not support the legacy names anymore.

charts/velero/templates/NOTES.txt
```diff
@@ -11,3 +11,55 @@ Once velero server is up and running you need the client before you can use it
...
+{{- $breaking_title := "\n" }}
+{{- $breaking_title = print $breaking_title "\n#################################################################################" }}
+{{- $breaking_title = print $breaking_title "\n######   BREAKING: The config values passed contained no longer accepted    #####" }}
+{{- $breaking_title = print $breaking_title "\n######             options. See the messages below for more details.        #####" }}
+{{- $breaking_title = print $breaking_title "\n######                                                                      #####" }}
+{{- $breaking_title = print $breaking_title "\n######             To verify your updated config is accepted, you can use   #####" }}
+{{- $breaking_title = print $breaking_title "\n######             the `helm template` command.                             #####" }}
+{{- $breaking_title = print $breaking_title "\n#################################################################################" }}
+
+{{- if hasKey .Values "resticTimeout" }}
```
qiuming-best, line 29: Names validation

charts/velero/templates/NOTES.txt
```diff
@@ -11,3 +11,55 @@ Once velero server is up and running you need the client before you can use it
...
+{{- if hasKey .Values "restic" }}
+{{- $breaking = print $breaking "\n\nREMOVED: restic has been removed, and it is named nodeAgent" }}
+{{- end }}
+
+{{- if hasKey .Values.configMaps "restic-restore-action-config" }}
+{{- $breaking = print $breaking "\n\nREMOVED: restic-restore-action-config has been removed, and it is named fs-restore-action-config" }}
+{{- end }}
+
+{{- range $key, $value := .Values.configMaps }}
+{{- eq $key "fs-restore-action-config" }}
```
qiuming-best, line 54: Here we couldn't use `{{- if hasKey index .Values.configMaps.fs-restore-action-config.labels "velero.io/restic" }}`, or it would be parsed failed with `templates/NOTES.txt:54): bad character U+002D '-'`, so I traversed the kv.

charts/velero/Chart.yaml
qiuming-best: Modified

jenting:
Since this PR introduced a bit of breaking backward compatibility because of the naming change.
We could consider fixing this issue #421 along with releasing velero v1.10.0 chart.

@qiuming-best @ywk253100 @reasonerjt @sseago @shubham-pampattiwar WDYT?

CharlieC3:
If pre-existing values config names are changing and the new CRDs are not backwards compatible, the chart's MAJOR version number should be rev'd to stay in alignment with semantic versioning expectations (e.g. v2.0.0. Either one of these reasons would be enough to justify this.

CharlieC3:
If pre-existing values config names are changing and the new CRDs are not backwards compatible, I think the chart's MAJOR version number should be rev'd to stay in alignment with semantic versioning expectations (e.g. v2.0.0). Either one of these reasons would be enough to justify this.
It doesn't need to match the same version as the application it deploys, and most Helm charts don't for this very reason.

This, along with ample documentation in the README, should be enough to inform users that the two deployments are incompatible.
I'd love to be able to use a chart supporting the 1.10.0 Velero release, and hope that separating the CRDs to a different chart would be something that can be addressed in another PR to get this one merged sooner :)

jcpunk:
Once this is merged, can it be submitted to artifiacthub.io?

jenting:
> Once this is merged, can it be submitted to artifiacthub.io?

Yes, but why?

jcpunk:
I tend to use artifacthub as my starting point for finding helm charts.  Having this posted there should help increase the visibility of this project.

sdelrio:
> > I tend to use artifacthub as my starting point for finding helm charts. Having this posted there should help increase the visibility of this project.
> 
> Create you create an issue for this request? Thank you.

It looks like it is already there:

https://artifacthub.io/packages/helm/vmware-tanzu/velero

charts/velero/values.yaml
```diff
@@ -335,6 +333,12 @@ configuration:
   # additional key/value pairs to be used as environment variables such as "AWS_CLUSTER_NAME: 'yourcluster.domain.tld'"
   extraEnvVars: {}
 
+  # Set true for backup all pod volumes without having to apply annotation on the pod when used file system backup Default: false. Other option: false.
```
jenting, line 336:
```suggestion
  # Set true for backup all pod volumes without having to apply annotation on the pod when used file system backup Default: false.
```

charts/velero/values.yaml
```diff
@@ -284,8 +284,12 @@ configuration:
   # here if using a non-default value. The `velero server` default values are shown in the
   # comments below.
   # --------------------
+  # `velero server` not empty. kopia or restic
```
jenting, line 287: > `velero server` not empty. kopia or restic

This is an optional field, correct? This sentence looks like we must set it as either kopia or resitc.

charts/velero/values.yaml
```diff
@@ -398,15 +402,15 @@ backupsEnabled: true
 # Whether to create volumesnapshotlocation crd, if false => disable snapshot feature
 snapshotsEnabled: true
 
-# Whether to deploy the restic daemonset.
-deployRestic: false
+# Whether to deploy the node-agent daemonset.
+defaultNodeAgent: false
```
jenting, line 406: Have you considered using the name `deployNodeAgent`?

jenting (commented):
@qiuming-best 
All good, jut some suggestions.

charts/velero/values.yaml
qiuming-best: yes, I've changed it to deployNodeAgent

charts/velero/values.yaml
qiuming-best: I've changed it to an optional field, if not set, the default value would be `restic`

charts/velero/values.yaml
qiuming-best: removed

jenting (approved):
LGTM

Thank you @qiuming-best

jenting:
@reasonerjt @ywk253100 PTAL

charts/velero/values.yaml
```diff
@@ -284,8 +284,12 @@ configuration:
   # here if using a non-default value. The `velero server` default values are shown in the
   # comments below.
   # --------------------
+  # `velero server` default: restic
```
reasonerjt, line 287: You don't have to fix this in this PR, but if these values are to be set by the user, we better add more explanations in the comment

reasonerjt (approved):
Have some trivial comments, but I think this PR looks great!

PR #425, opened 2022-12-07
[velero] use value instead of key for extraEnvVars

k4mrul:
Signed-off-by: k4mrul <kamrulahsan06@gmail.com>

#### Special notes for your reviewer: Use value instead of key for `credentials.extraEnvVars`
```
credentials:
  extraEnvVars: 
    DIGITALOCEAN_TOKEN: do_token  <-- Currently this value is ignored.
```

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[velero]`)

PR #428, opened 2022-12-23
[velero] fix: secret credentials key

ydkn:
Use `key` as `key` for `secretRef` instead of `value`.

Error on current version when using `credentials.extraEnvVars`:
```
Helm release "velero/velero" failed to initialize completely. Use Helm CLI to investigate.: failed to become available within allocated timeout. Error: Helm Release velero/velero: cannot patch "velero" with kind Deployment: Deployment.apps "velero" is invalid: spec.template.spec.containers[0].env[6].valueFrom.secretKeyRef.key: Invalid value: "<e.g. AWS secret access key>": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/templates/deployment.yaml
+++ b/charts/velero/templates/deployment.yaml
@@ -218,7 +218,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "velero.secretName" $ }}
-                  key: {{ default "none" $value }}
+                  key: {{ default "none" $key }}
           {{- end }}
           {{- end }}
           {{- if .Values.lifecycle }}
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.10.0
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 3.0.0
+version: 3.0.1
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:


charts/velero/templates/deployment.yaml
```diff
@@ -218,7 +218,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "velero.secretName" $ }}
-                  key: {{ default "none" $value }}
+                  key: {{ default "none" $key }}
```
jenting, line 221: @k4mrul Can you help check this? Because it's the revert of your change #425

k4mrul: `a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')`

Possibly, the key name is wrong. Here is the example of usage

`secret.yaml`
```
apiVersion: v1
kind: Secret
metadata:
    name: velero-dospace-secret
    namespace: velero
stringData:
    dovalue: <do-token>
    cloud: |-
        [default]
        aws_access_key_id=<key>
        aws_secret_access_key=<secret>
```

`values.yaml`
```
credentials:
  existingSecret: velero-dospace-secret
  secretContents: {}

  extraEnvVars: 
    DIGITALOCEAN_TOKEN: dovalue
```

Now if we apply secrets.yaml and use the above values.yaml in velero, the pod will mount key and value from secret.

```
    - name: DIGITALOCEAN_TOKEN
      valueFrom:
        secretKeyRef:
          key: dovalue
          name: velero-dospace-secret
```

I just tested this config with chart verion 3.0.0. Working fine.

mhumeSF: There should be a note that the secret key has to be named `cloud`?

ydkn: Thanks @k4mrul for the clarification. should have read the updated values.yaml more carefully as I was expecting `credentials.extraEnvVars` behaves the same as e.g. `nodeAgent.extraEnvVars`.

PR #450, opened 2023-04-07
[velero]: Switch order of init & main in update-crds job

tuananh:

right now, we're copying sh and kubectl over to velero image but doing this required that kubectl image and velero image are same OS (shared libs, etc..)

I'm re-ordering velero to be init container and will generate CRD yaml to /tmp.

main container will be kubectl and only apply the generated yaml file in `/tmp`. This way, we can say use kubectl alpine version built with musl and velero image can be Debian based for example.

use case:

- bitnami/kubectl has lots of CVE and we want to switch to sth else.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: 1.10.3
+appVersion: 1.10.2
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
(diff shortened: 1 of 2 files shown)


tuananh:
@jenting my bad. this doesnt' work because velero install command cannot generate file to `/tmp` without a shell, unless `velero` cli support sth like `velero install --crds-only --dry-run -o yaml --output-file /tmp/crds.yaml`

PR #468, opened 2023-06-14
Default to empty volumeSnapshotLocation.

toothbrush:
#### Checklist
- [ ] Chart Version bumped
- [ ] Title of the PR starts with chart name (e.g. `[velero]`)


jenting:
@toothbrush Thanks for your contribution.

Perhaps you could change `snapshotsEnabled: false` in your values.yaml, without having to make code change.
Please let me know if it works for you or not.

toothbrush:
> Perhaps you could change `snapshotsEnabled: false` in your values.yaml, without having to make code change.
> Please let me know if it works for you or not.

That's actually what i ended up doing, after staring more at the Helm charts 😃 

However i still believe it's a bug that in the default configuration, this Helm chart (appears to) result in broken config (i.e, an invalid entry in the `volumeSnapshotLocation` slice).

jenting (changes requested):
Comment left

PR #481, opened 2023-07-27
[velero] create prometheusRule only if capabilities are present

nobbs:

Similar to existing `podMonitor` and `serviceMonitor` templates, only create the `prometheusRule` if the capabilities are present in the cluster (and rules are defined).

#### Checklist

- [x] Chart Version bumped


charts/velero/templates/prometheusrule.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.metrics.enabled .Values.metrics.prometheusRule.enabled }}
+{{- if and (and .Values.metrics.enabled .Values.metrics.prometheusRule.enabled) (or (not .Values.metrics.prometheusRule.autodetect) (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1")) (.Values.metrics.prometheusRule.spec) }}
```
jenting, line 1: Is the (.Values.metrics.prometheusRule.spec) purpose to check the .Values.metrics.prometheusRule.spec exists, not nil?

nobbs: Yes, exactly. If it's nil / an empty dict the rule would have been created but without any actual rules inside. This way it will only create if enabled, capabilities are met and some rule specs are defined.

jenting (approved):

PR #509, opened 2023-10-16
fix(velero): remove useless `nodeAgent.privileged`

cwrau:
#### Checklist
- [ ] Chart Version bumped


jenting:
@qiuming-best could you help review this?

charts/velero/README.md
```diff
@@ -90,6 +90,10 @@ helm upgrade vmware-tanzu/velero <RELEASE NAME> --reuse-values --set configurati
 
 ## Upgrading
 
+### Upgrading to 5.1.1
+
+This version removes the `nodeAgent.privileged` field, you should use `nodeAgent.containerSecurityContext.privileged` instead
+
```
jenting, line 96: This is the note for velero major/minor version upgrade doc.
Please consider have a comment in values.yaml instead.

jenting (changes requested):

charts/velero/README.md
jenting: Generally, we made backward compatible for each helm chart minor/patch version.
If it can't be backward compatible, we bump the helm chart major version.

jenting: > Yes, I understand, we're doing the same for our charts, but where do you document breaking changes or little migration steps like this?

Sorry, no. We don't have the doc sections mentioned the steps for the breaking change.
If you like, we could create a new doc section for the breaking change one.
But for this PR case, I think it should able to make the helm chart backward compatible.

cwrau: Yeah, I was thinking about the "major break"ability of this change as well...

The current implementation, just removing the field, isn't really backwards compatible.

For one there is no schema, so the user doesn't even get a validation error, and on the otherside the field is just silently ignored as if it would still work.

So this should either be a breaking change or I'd have to merge the privileged field.

Third possibility would be to add a validation to disallow this field in lieu of implementing the whole schema for this little change. But I would kinda classify this as a breaking change as well.

Fourth solution would be to just silently drop it but also not mark it as a breaking change, which wouldn't be very nice to the users tho 😅

What do you think?

jenting: Or let's keep it as is since the purpose it to deduplicate the duplicated fields. WDYT?

cwrau: > Or let's keep it as is since the purpose it to deduplicate the duplicated fields. WDYT?

Yeah, then I'll just change this to a major release as well. See the README on how I thought we could add info about this 👌

jenting:
Could you help resolve the conflicting file? Thanks.

jenting:
> Could you help resolve the conflicting file? Thanks.

@cwrau do you have time to proceed this issue?

charts/velero/README.md
```diff
@@ -87,8 +87,13 @@ If a value needs to be added or changed, you may do so with the `upgrade` comman
 ```bash
 helm upgrade vmware-tanzu/velero <RELEASE NAME> --reuse-values --set configuration.backupStorageLocation[0].provider=<NEW PROVIDER>
 ```
+## Upgrading Chart
 
-## Upgrading
+### Upgrading to 6.0.0
+
+This version removes the `nodeAgent.privileged` field, you should use `nodeAgent.containerSecurityContext.privileged` instead
```
jenting, line 94: Do you think we should create another doc, such as [BREAKING_CHANGES.md](https://github.com/elastic/helm-charts/blob/main/BREAKING_CHANGES.md) to indicate which versions break the changes rather than write it down in the README.md?

cwrau: I personally prefer to have it all in the README so people don't miss it or have to click on another link to get to critical information.

jenting: What do you think @qiuming-best @ywk253100 @reasonerjt

qiuming-best: There will be more and more breaking changes with time, and I wish we had a `BREAKING_CHANGE.md` file to declare these centrally. We then link it  in README.md

qiuming-best (approved):

PR #561, opened 2024-03-31
modify the default value of server

muxuelanKK:
modify the default value of server
image


charts/velero/values.yaml
```diff
@@ -391,17 +391,17 @@ configuration:
   # here if using a non-default value. The `velero server` default values are shown in the
   # comments below.
   # --------------------
-  # `velero server` default: kopia
+  # `velero server` default: restic
```
qiuming-best, line 394: The default uploader should still be `Kopia`.
Although our velero source codes the default value as `restic`, but we specify the default value through `velero install`, in which the default value is `Kopia`, and the value would pass to `velero server`.
And we will change the default value for `velero server` in the future.

Issue #554, opened 2024-02-29
Adhere to semver


cwrau:
**Describe the problem/challenge you have**
Currently the velero helm chart kinda randomly decides which part of the version gets bumped when a change is merged, see https://github.com/vmware-tanzu/helm-charts/issues/549#issuecomment-1958387077. This means that every user has to look at every little (patch/minor) release to see if something broke for them.

**Describe the solution you'd like**
Adhere to [semver](https://semver.org).

**Anything else you would like to add:**
This helps users judge releases just by their version number and use tools to auto-update or to automatically create small PRs to update versions.

qiuming-best:
Our chart would release for two cases:
  - Velero release
  - Chat itself updates

What about we make a rule like below:
  
    Supposing the Velero version is X.Y.Z
    X.Y determined the Major version of the chart
    Such as 6 -> 1.13; 7 ->1.14;  8 -> 1.15
    Z determined the Minor version of the chart
    Such as 0 -> 1.13.0; 1->1.13.1; 2 -> 1.13.2
    The patch version of the chart is determined by the chart itself.

Such as if Velero release V1.13.1
the first chart version for it would be 6.1.0

If we have one urgent breaking change, it will break the semver rule.

we could add a table in the README to record the relationship between chart versions and Velero versions. 
We would still follow the aforementioned rules as a base. However, if a breaking change occurs, we would upgrade the major version and continue to follow the same rules. For example, initially:
6 -> 1.13; 7 ->1.14;
Then, if version 7 experiences a breaking change, we would upgrade the major version. At this point, it would be:
6 -> 1.13; 7 ->1.14; breaking change occurred 8 ->1.14;

qiuming-best:
Any ideas about it? @jenting @reasonerjt @ywk253100

cwrau:
> Our chart would release for two cases:
>   - Velero release
>   - Chat itself updates
> 
> What about we make a rule like below:
>   
>     Supposing the Velero version is X.Y.Z
>     X.Y determined the Major version of the chart
>     Such as 6 -> 1.13; 7 ->1.14;  8 -> 1.15
>     Z determined the Minor version of the chart
>     Such as 0 -> 1.13.0; 1->1.13.1; 2 -> 1.13.2
>     The patch version of the chart is determined by the chart itself.
> 
> Such as if Velero release V1.13.1
> the first chart version for it would be 6.1.0
> 
> If we have one urgent breaking change, it will break the semver rule.
> 
[...]

PR #569, opened 2024-04-19
[velero] Adhere chart to semver

qiuming-best:
#### Checklist
- [ ] Chart Version bumped

Fix https://github.com/vmware-tanzu/helm-charts/issues/554

- Add helm release instruction document.
- Modify release workflow, make any commit merged into `main` or `velero-helm-charts-v*` trigger helm chart release.
- Modify lint-test, trigger test if the current submit different from the target branch

### DIFF
--- a/RELEASE-INSTRUCT.md
+++ b/RELEASE-INSTRUCT.md
@@ -1,0 +1,39 @@
+Here is an instruction about the Velero helm charts release.
+
+It involves Helm charts version control and branch control.
+
+### Helm Charts Version Control
+
+Our Helm charts are released under two scenarios:
+- **Velero Release:** When Velero itself is released.
+
+- **Chart Updates:** When there are updates specific to the Helm charts.
+
+The version is defined in [Chart.yaml](charts/velero/Chart.yaml).
+
+#### Guidelines
+To comply with the [Semantic Versioning (semver)](https://semver.org/#summary) rule, follow these instructions:
+- **Velero Release:**
+  - For each Velero minor version release, increase the major version of the Helm charts. For example, if Velero v1.13.0 is released, the Helm charts version becomes 6.0.0; for Velero v1.14.0, it becomes 7.0.0.
+  
+  - For each Velero patch version release, increase the minor version. For instance, for Velero v1.13.1, the Helm charts version is 6.1.0; for Velero v1.13.2, it's 6.2.0.
+- **Helm Charts Updates:**
+  - For breaking changes, increase the major version.
+  
+  - For added functionality, increase the minor version.
+  
+  - For bug fixes, increase the patch version.
+
+#### Note:
+Breaking changes are only allowed for the **latest Helm version**. This restriction ensures that older Helm versions do not have higher major versions than newer ones, preventing helm upgrade issues.
+
+### Branch Control
+
+Follow these rules for branch control:
+- Our main branch should always align with the latest Velero major version. For example, if Velero v1.13 is the latest major version, any updates in the main branch should target Velero v1.13.
+
+- Create a new branch prefixed with `velero-helm-charts-v` when Velero releases a new major version. For instance, if the main branch refers to Velero v1.13 and Velero v1.14 is released, name the new branch `velero-helm-charts-v7`. The branch version should match the major version of the Helm charts.
+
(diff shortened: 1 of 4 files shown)


siegenthalerroger:
looks good to me :)

RELEASE-INSTRUCT.md
```diff
@@ -0,0 +1,39 @@
...
+- **Velero Release:** When Velero itself is released.
+
+- **Chart Updates:** When there are updates specific to the Helm charts.
+
+The version is defined in [Chart.yaml](charts/velero/Chart.yaml).
+
+#### Guidelines
+To comply with the [Semantic Versioning (semver)](https://semver.org/#summary) rule, follow these instructions:
+- **Velero Release:**
+  - For each Velero minor version release, increase the major version of the Helm charts. For example, if Velero v1.13.0 is released, the Helm charts version becomes 6.0.0; for Velero v1.14.0, it becomes 7.0.0.
```
jenting, line 17: Personally, I don't think the helm chart itself follows the semver because the Velero bumps minor version (functionality change) but the corresponding helm chart changes the major version (breaking changes).

We could consider maintains the velero helm chart and Velero itself with this [format ](https://github.com/rancher/charts?tab=readme-ov-file#upstream-charts). This makes the helm chart follows the semver as well, also we can easily check the helm chart upstream Velero version.

jenting (commented):
@qiuming-best  Thank you for coordinate the topic and comes out the idea and documentation. I have another ideas on this one, please take a look of my comments.

RELEASE-INSTRUCT.md
qiuming-best: @jenting 
I do not quite understand the format :
`The upstream charts follow this versioning: 1.0.#+upX.Y.Z`
Could you please give more details explanation?

Our main goal for this PR about versioning is to solve the  [conflicts in chart versions problem](https://github.com/vmware-tanzu/helm-charts/issues/554#issuecomment-1985113595)
Not sure the format you mentioned could avoid the conflicts in chart versions problem?

siegenthalerroger: @jenting the reasoning behind the chart *requiring* a major bump for every minor velero bump is because velero itself doesn't follow semver. There are often changes in minor velero updates that require changes to be made to the deployment. To a certain extent the helm chart could catch these and make them non-breaking, however the decision here would be to forego that effort and just admit that the helm chart just "moves fast and breaks things" in a sense.

siegenthalerroger: Having looked at the rancher example, they actually have the same "ruleset" in a sense. The helm major version is always bumped for a minor rancher release (the `100.x.y` examples), just with an additional `+up...`. 

Personally I don't think the additional `+up` is necessary, that's what the `appVersion` field in the `Chart.yaml` is for, and it's what we use to track the upstream version in our deployments.

jenting: What if the Velero bump the major release? How should we tackle in the Velero helm chart?

My thought was
- if the Velero bump patch/minor/major version, the helm chart itself should bump patch/minor/major version.
- if the helm chart itself have to fix/feature/breaking change, bump the helm chart version <major>.<minor>.<patch>-<Velero-version>. For example: current version is `6.1.0-v1.13.0`
  - The helm chart bug fix would be `6.1.1-v1.13.0`
  - The helm chart feature enhance would be `6.2.0-v1.13.0`
  - The helm chart breaking change would be `7.0.0-v1.13.0`
  - The velero bug fix would be `6.1.1-v1.13.1`
  - The velero feature enhance would be `6.2.0-v1.14.0`
  - The velero bbreaking change would be `7.0.0-v2.0.0`

siegenthalerroger: To the first question: just do a major bump as you'd expect.

To your suggestion: While that _sounds_ fine, the issue raised was that the helm chart can sometimes manage to update to new velero minor versions without a breaking change, but sometimes it does. IMO, while the originally suggested change veers on the safe side (always doing major bumps for minor velero updates, just incase they have a breaking change), you could cover it just by making sure it major bumps if a breaking change is required.

jenting (approved):

ywk253100 (approved):

PR #623, opened 2024-10-23
[velero] fix: consistent node-agent podLabels

foyerunix:
Hello,

The node-agent was using ".Values.podLabels" in the DaemonSet while using ".Values.nodeAgent.labels" in the podMonitor.

This led to an inconsistency and made it impossible to have different labels for the node-agent compared to the Velero deployment.

Best Regards.

### DIFF
--- a/charts/velero/values.yaml
+++ b/charts/velero/values.yaml
@@ -557,6 +557,10 @@ nodeAgent:
   # labels to set for the node-agent daemonset. Optional.
   labels: {}
 
+  # Additional pod labels for the node-agent daemonset. Optional
+  # ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+  podLabels: {}
+
   # will map /scratch to emptyDir. Set to false and specify your own volume
   # via extraVolumes and extraVolumeMounts that maps to /scratch
   # if you don't want to use emptyDir.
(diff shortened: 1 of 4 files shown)


charts/velero/values.yaml
```diff
@@ -557,6 +557,10 @@ nodeAgent:
   # labels to set for the node-agent daemonset. Optional.
   labels: {}
 
+  # Additional pod labels for the node-agent daemonset. Optional
+  # ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+  podLabels: {}
```
jenting, line 562: Isn't there is the label for the node-agent?
https://github.com/vmware-tanzu/helm-charts/blob/f0f07defa8273b0493806032349539f948da40b4/charts/velero/values.yaml#L557-L558

foyerunix: Hello,

Yes this is where the labels for the node-agent DaemonSet are defined. Meanwhile the PodMonitor will not match the labels of the DaemonSet but the labels of the pods created by the DaemonSet.

The labels of the pods of the node-agent DaemonSet are defined as ".Values.podLabels", hence they will be the same as the labels of the Velero Deployment. But the PodMonitor will try to match the labels defined in ".Values.nodeAgent.labels".

So with the current chart it is impossible to have different labels between the Velero Deployment and the Velero node-agent.

Best Regards.

jenting:
@foyerunix sorry for the late reply. Could you please bump the chart version? The others LGTM. Thanks.

foyerunix:
Hello,

Chart version bumped.

Best Regards.

jenting (approved):

blackpiglet (approved):

PR #699, opened 2025-07-28
[velero] Update example PrometheusRule

mc2285:
The provided example shall never be one that gives false negatives on the subject of whether a backup is failing. The consequences may be catastrophic if someone does not notice it is an absolutely horrible example that is giving a false sense of security.

It would be better to remove the example altogether than to provide one that can keep claiming everything is in order when there is a chain of failed backups instead.

If it is impossible to include mine / another one that actually works, please remove the bad example that the helm chart's `values.yaml` currently provides

#### Checklist
- [x] Chart Version bumped, please refer to the [chart version instruction](https://github.com/vmware-tanzu/helm-charts/blob/main/RELEASE-INSTRUCT.md#guidelines) -> N/A, updated code that is commented out
- [x] Variables are documented in the values.yaml or README.md -> N/A, updated code that is commented out

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.16.1
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 10.0.10
+version: 10.0.11
 home: https://github.com/vmware-tanzu/velero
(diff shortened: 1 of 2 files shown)


jenting:
Please bump the Chart version

jenting (approved):

blackpiglet (approved):

Issue #727, opened 2026-02-03
upgradeCRDs not able to pull the image on Kubernetes 1.35


flmmartins:
**What steps did you take and what happened:**

I upgraded my kubernetes to 1.35 lately and  I use default values for my Velero and today when I deploy an update on memory limits I got:

message: 'Back-off pulling image "docker.io/bitnamilegacy/kubectl:1.35": ErrImagePull: rpc error: code = NotFound desc = failed to pull and unpack image "docker.io/bitnamilegacy/kubectl:1.35":  failed to resolve reference "docker.io/bitnamilegacy/kubectl:1.35": docker.io/bitnamilegacy/kubectl:1.35: not found'

If I check https://hub.docker.com/r/bitnamilegacy/kubectl/tags this repository no longer exists and it only contains kubectl:1.33.

I assume because I upgraded Kubernetes it wants to use 1.35 which is my kubernetes version but the image no longer exists

**What did you expect to happen:**

The upgrade job to run with no trouble as always.

**Environment:**

- helm version (use `helm version`): 4.0.5
- helm chart version and app version (use `helm list -n <YOUR NAMESPACE>`):
Chart: velero-11.3
App: 1.17.1
- Kubernetes version (use `kubectl version`): 1.35.0
- Kubernetes installer & version:1.35.0
- Cloud provider or hardware configuration: On premise

jenting:
Latest helm chart release, please help verify it. Thank you.

flmmartins:
Hello @jenting ! It worked! Thank you so muchhh!!

PR #706, opened 2025-09-04
[velero] Remove bitnami nicely by inverting the original idea

albundy83:
#### Special notes for your reviewer:

Try to fix https://github.com/vmware-tanzu/helm-charts/issues/698
#### Checklist
- [x] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [x] Chart Version bumped, please refer to the [chart version instruction](https://github.com/vmware-tanzu/helm-charts/blob/main/RELEASE-INSTRUCT.md#guidelines)
- [x] Variables are documented in the values.yaml or README.md
- [x] Title of the PR starts with chart name (e.g. `[velero]`)

For moment, it's just an idea to remove bitnami kubectl and the need for sh :)

What do you think ?

Issue #698, opened 2025-07-21
Bitnami kubectl image change – migration plan for Velero helm chart


a-pallari:
**Describe the problem/challenge you have**
The Velero Helm chart pulls the bitnami/kubectl image https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L316-318. Bitnami has announced that its public container catalog will no longer be maintained https://github.com/bitnami/charts/issues/35164.

**Describe the solution you'd like**
Is there a migration plan to an actively maintained kubectl image?

**Anything else you would like to add:**
This deprecation might turn out to be a non-issue, but I wanted to flag it early in case action is needed.

SheryarButt:
@rajivml yes they require subscription, that's the whole reason for this shift 💵

nemobis:
>  unless there's a viable alternative kubectl image with which, sh, and kubectl

The image [eclipsefdn/kubectl](https://hub.docker.com/r/eclipsefdn/kubectl) should do.

kaovilai:
How about https://github.com/alpine-docker/k8s

kaovilai:
From my reading bitnamisecure is free on the latest tag, why can't we use it?

SerQuicky:
I just tried both of latest bitname images (bitnami/kubectl and bitnamisecure/kubectl) and interestingly the bitnamesecure image did not work with the crd process. Did anyone also try it out and what was the result?

albundy83:
I have made this pull request https://github.com/vmware-tanzu/helm-charts/pull/706 to remove the need for Bitnami kubectl and shell.

The idea is to first generate the crds yaml and then to apply them using using kubectl, so you don't need shell :)

kaovilai:
I like that @albundy83

albundy83:
Yes : https://hub.docker.com/r/bitnamilegacy/kubectl/tags?name=1.34

But if we don't do anything, it will be break for [1.34](https://hub.docker.com/r/bitnami/kubectl/tags?name=1.34) and for all others releases after September the 29th (see [here](https://community.broadcom.com/tanzu/blogs/beltran-rueda-borrego/2025/08/18/how-to-prepare-for-the-bitnami-changes-coming-soon)).

SheryarButt:
I suggest locking the image tag in the chart to [1.33.4](https://hub.docker.com/r/bitnami/kubectl/tags?name=1.33.4), as it's the latest one available, until we have a proper fix.

[...]

PR #728, opened 2026-02-04
fix(velero): pin kubectl image to official registry for k8s 1.34 and 1.35

its-nihal:
Velero upgrade/cleanup jobs were failing on Kubernetes 1.34 and 1.35 due to the default kubectl image resolving to docker.io/bitnamilegacy/kubectl:1.34, and docker.io/bitnamilegacy/kubectl:1.35, which does not exist. The Bitnami legacy kubectl repository is frozen and only publishes images up to Kubernetes 1.33.

This change explicitly pins the kubectl image to the official Kubernetes registry (registry.k8s.io/kubectl) with a compatible version, preventing ErrImagePull failures and making the deployment resilient to future Kubernetes upgrades.

No functional behavior change beyond fixing image pull failures.

Issue raised - https://github.com/vmware-tanzu/helm-charts/issues/727

#### Checklist
- [ ] Chart Version bumped, please refer to the [chart version instruction](https://github.com/vmware-tanzu/helm-charts/blob/main/RELEASE-INSTRUCT.md#guidelines)
- [ ] Title of the PR starts with chart name (e.g. `[velero]`)

### DIFF
--- a/charts/velero/templates/_helpers.tpl
+++ b/charts/velero/templates/_helpers.tpl
@@ -116,3 +116,16 @@ Calculate the checksum of the credentials secret.
 {{- define "chart.config-checksum" -}}
 {{- tpl (print .Values.credentials.secretContents .Values.credentials.extraEnvVars ) $ | sha256sum -}}
 {{- end -}}
+{{/*
+Kubectl image tag
+- If user explicitly sets kubectl.image.tag → use as-is
+- Otherwise infer from Kubernetes version and prefix with 'v'
+- Use Major.Minor inferred version and append '.0'
+*/}}
+{{- define "velero.kubectlImageTag" -}}
+{{- if .Values.kubectl.image.tag -}}
+{{- .Values.kubectl.image.tag -}}
+{{- else -}}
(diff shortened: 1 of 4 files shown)


mlec1:
I might be wrong but as explained here https://github.com/vmware-tanzu/helm-charts/issues/698, the official kubectl image is not a good fit because it doesn't have a shell which is required in the the cleanup job.

markbluhm:
If we're still okay with using bitnami via their elastic container registry (public.ecr), we ought to use that instead... but I'm a much bigger fan of not using bitnami.

So I'd say, despite this not addressing clean up scripts, it at least addresses part of the problem. I'd be a fan of seeing this planned for an upcoming release but replaced by another solution that handles the clean up job either in that release or in a follow up release.

Not normally one to encourage a "half-fix" but half is better than 0% and it gets a lot of people back on default values instead of overrides.

markbluhm (approved):

jenting:
I think the PR #706 had addressed this issue so this PR may consider close it.

PR #740, opened 2026-06-18
[velero] Add optional PodDisruptionBudget

somaz94:
velero runs a single replica by default, so this PDB is opt-in (`podDisruptionBudget.enabled: false`). It's handy once you scale to multiple replicas and want to keep a pod up during node drains and upgrades.

Supports `minAvailable`/`maxUnavailable`, an optional `unhealthyPodEvictionPolicy`, and extra annotations/labels. Bumped the chart to 12.1.0 and enabled it in `ci/test-values.yaml` so install/upgrade actually exercises it.

Ran `ct lint` and `ct install --upgrade` locally on kind, both pass.

- [x] DCO signed
- [x] Chart version bumped (12.0.3 -> 12.1.0)
- [x] Variables documented in the values.yaml
- [x] Title starts with `[velero]`

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.18.1
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 12.0.3
+version: 12.1.0
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/ci/test-values.yaml
+++ b/charts/velero/ci/test-values.yaml
@@ -43,6 +43,11 @@ serviceAccount:
   server:
     name: velero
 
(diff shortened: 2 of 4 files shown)


somaz94:
Pushed a follow-up fix (4de5996) for the failing `lint-test` matrix.

The failure was in the **upgrade** test, not install: `ct install --upgrade` reuses the previously-installed release's values, so the newly-added `podDisruptionBudget` key isn't present and `.Values.podDisruptionBudget` is nil during the upgrade render — `nil pointer evaluating interface {}.enabled`.

Made the guard nil-safe, matching the convention already used in this chart (e.g. `(.Values.upgradeCRDsJob).extraVolumeMounts` in `upgrade-crds.yaml`):

```diff
-{{- if .Values.podDisruptionBudget.enabled }}
+{{- if (.Values.podDisruptionBudget).enabled }}
```

Verified locally: nil case renders nothing without error, `enabled=true` renders the PDB, and `helm lint` passes. The new CI run is in `action_required` — could a maintainer approve the workflow run? Thanks!

jenting (approved):

blackpiglet (approved):

PR #743, opened 2026-06-29
[velero] Support deploying plugins using Kubernetes Image Volumes

### LINKED ISSUE #742: [Feature] Simplify plugins deployment using Kubernetes Image Volumes
**Describe the problem/challenge you have**

The chart deploys Velero plugins by copying their binaries into a shared `emptyDir` volume using `initContainers`. This has three drawbacks:

1. **Startup time** — init containers run sequentially before the Velero container starts.
2. **Resource management** — each plugin needs its own CPU/memory tuning for a short-lived container, which complicates quota and autoscaling.
3. **Distroless images** — modern plugin images ship without a shell or `cp`, so extracting the binary requires workarounds.

**Describe the solution you'd like**

Use Kubernetes [Image Volumes](https://kubernetes.io/docs/concepts/storage/volumes/#image) (beta and on by default since Kubernetes 1.33, behind the `ImageVolume` feature gate since 1.31) to mount each plugin OCI image directly as a read-only volume, without execution.

I propose adding a new, simplified configuration block in the chart's `values.yaml`:

```yaml
plugins:
  - image: velero/velero-plugin-for-aws:v1.10.0
  - image: velero/velero-plugin-for-csi:v0.7.1
```

The Helm chart would then automatically loop through these and translate them into the appropriate volumes (using the `image:` type) and `volumeMounts` (using `subPath` to extract the binary into `/plugins/`) within the Velero Deployment and DaemonSet templates.

- `plugins` would be opt-in and combinable with `initContainers`; existing behaviour is unchanged when it is empty.
- The current workaround is to set `extraVolumes` / `extraVolumeMounts` manually, which works but exposes a lot of Kubernetes plumbing to chart users. A dedicated `plugins` value would abstract this elegantly and make the chart modern-ready.
- Originally filed as velero-io/velero#9953 before realizing the chart lives in this repo.

**Environment:**

- helm version (use `helm version`): v3.19.0
- helm chart version and app version (use `helm list -n <YOUR NAMESPACE>`): velero chart 12.1.0, app v1.18.1
- Kubernetes version (use `kubectl version`): v1.35.x
- Kubernetes installer & version: FluxCD-managed cluster

fredgate:
**Which issue(s) this PR fixes**:

Closes #742

**What this PR does / why we need it**:

Adds an optional `plugins` value to deploy Velero plugins using Kubernetes [Image Volumes](https://kubernetes.io/docs/concepts/storage/volumes/#image), as an alternative to `initContainers`:

```yaml
plugins:
  - image: velero/velero-plugin-for-aws:v1.13.1
  - image: velero/velero-plugin-for-csi:v0.7.1
```

For each entry the chart generates an `image` volume and mounts the plugin binary read-only under the plugin directory (`/plugins/<name>`). By default the binary is taken from `plugins/<name>` inside the image (the convention used by the Velero plugin images); `subPath` can be overridden per plugin when the binary lives elsewhere, or pointed at a directory to expose several binaries.

Only the plugin binary is exposed (via `subPath`), so unrelated executables in the plugin image are never picked up by Velero's recursive plugin scan.

- Opt-in and fully backward compatible: when `plugins` is empty the rendered manifest is unchanged, and `initContainers` keeps working. `plugins` and `initContainers` can be combined.
- Requires the `ImageVolume` feature with `subPath` support: Kubernetes >= 1.33 (beta, enabled by default) and a runtime such as containerd >= 2.1 / CRI-O >= 1.33.
- Only the Velero server Deployment is affected; the node-agent DaemonSet does not load Velero plugins.
- Pods that Velero creates dynamically are not broken by this change. The repository maintenance Job copies the server's `volumes`/`volumeMounts` (but not its `initContainers`): with `initContainers` the inherited `plugins` `emptyDir` arrives empty, whereas an image volume is self-populating, so the maintenance pod simply gets a harmless read-only mount of the plugin binary. It needs no plugins either way (kopia accesses storage natively via the BSL). Data mover "exposer" pods do not inherit the server volumes and load no plugins.

#### Checklist
- [x] Chart Version bumped, please refer to the [chart version instruction](https://github.com/vmware-tanzu/helm-charts/blob/main/RELEASE-INSTRUCT.md#guidelines)

### DIFF
--- a/charts/velero/Chart.yaml
+++ b/charts/velero/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.18.1
 kubeVersion: ">=1.16.0-0"
 description: A Helm chart for velero
 name: velero
-version: 12.1.0
+version: 12.2.0
 home: https://github.com/vmware-tanzu/velero
 icon: https://cdn-images-1.medium.com/max/1600/1*-9mb3AKnKdcL_QD3CMnthQ.png
 sources:
--- a/charts/velero/README.md
+++ b/charts/velero/README.md
@@ -80,6 +80,32 @@ If a value needs to be added or changed, you may do so with the `upgrade` comman
 helm upgrade <RELEASE NAME> vmware-tanzu/velero --namespace <YOUR NAMESPACE> --reuse-values --set configuration.backupStorageLocation[0].provider=<NEW PROVIDER>
 ```
 
+##### Deploying plugins with Image Volumes
+
+Instead of copying each plugin binary with an init container, plugins can be
+mounted directly from their OCI image using [Image Volumes](https://kubernetes.io/docs/concepts/storage/volumes/#image)
+via the `plugins` value:
+
+```yaml
+plugins:
+  - image: velero/velero-plugin-for-aws:v1.13.1
+  - image: velero/velero-plugin-for-csi:v0.7.1
+```
+
+The plugin binary is mounted read-only under the plugin directory, where Velero
+discovers it. By default the binary is taken from `plugins/<name>` inside the
+image (the convention used by the Velero plugin images); override `subPath` per
+plugin if the binary lives elsewhere, or point it at a directory to expose
+several binaries from one image. Compared to `initContainers`, this avoids the
+sequential init container startup, removes the per-plugin CPU/memory tuning,
+and works with distroless plugin images (no `cp` or shell required).
+
+This requires the `ImageVolume` feature with `subPath` support, i.e. Kubernetes
+>= 1.33 (beta, enabled by default), and a container runtime that supports image
+volumes (e.g. containerd >= 2.1, CRI-O >= 1.33). `plugins` and `initContainers`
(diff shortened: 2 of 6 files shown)


jenting (commented):
Shall we check the runtime k8s version above or equal to 1.33?

fredgate:
Good point — done in 3ccadeb. When `plugins` is set, the chart now fails rendering with a clear message if the target cluster is < 1.33: 
> velero: `plugins` mounts plugin binaries from image volumes via subPath,
> which requires Kubernetes >= 1.33, but the target cluster is v1.32.0.
> Use `initContainers` instead, or pass --kube-version when rendering offline.

The check uses `semverCompare ">=1.33.0-0" .Capabilities.KubeVersion.Version` and only triggers when `plugins` is non-empty, so it has no effect on existing deployments or when rendering offline without the feature (`--kube-version` can override the detected version).

jenting (approved):

